PR #3, opened 2019-04-13
Refactor and docs

scalvert:
Refactor of

- waiters.ts -> waiter-manager.ts
- simple-waiter.ts -> test-waiter.ts
- promise-waiter.ts
- tests
- docs

TODO:
- [x] Add tests for `debugInfo`
- [x] Add tests for custom waiters
- [x] Add test to ensure waiters aren't added with the same name
- [x] Add tests for `waitForPromise`
- [x] Correctly export types for public consumption
- [x] Determine public/private exports
- [x] Guard all executions with a `DEBUG` check, where appropriate


API.md
```diff
@@ -0,0 +1,186 @@
+<!-- Generated by documentation.js. Update this documentation by updating the source code. -->
+
+### Table of Contents
+
+- [Test Waiter Manager][1]
+  - [register][2]
+    - [Parameters][3]
```
rwjblue: Can we follow suite with @ember/test-helpers here and use the TOC depth option? The `Parameters` shenanigans in the table of contents drives me nuts!

README.md
```diff
@@ -1,37 +1,71 @@
-ember-test-waiters
-==============================================================================
+# ember-test-waiters
 
-[Short description of the addon.]
+This addon provides APIs to allow Ember testing to play nicely with other asynchronous
```
rwjblue:
```suggestion
This addon provides APIs to allow [@ember/test-helpers](https://github.com/emberjs/ember-test-helpers/) to play nicely with other asynchronous
```

README.md
```diff
@@ -1,37 +1,71 @@
...
+events, such as an application that is waiting for a CSS3
+transition or an IndexDB transaction. Waiters runs periodically
+after each async helper (i.e. `click`, `andThen`, `visit`, etc) has executed,
+until a predetermined condition is met. After the waiters finish, the next async helper
```
rwjblue: Hmm, this isn't _quite_ correct. Can we rephrase to make it clearer that the helpers themselves return a promise, and the "settledness" of these test waiters is part of the condition that determines when that promise resolves?

README.md
```diff
@@ -1,37 +1,71 @@
...
-------------------------------------------------------------------------------
+## Installation
 
 ```
 ember install ember-test-waiters
 ```
 
+## Usage
+
+Ember test waiters uses a minimal API to provide waiting functionality. This minimal API can be composed to accomodate various complex scenarios.
```
rwjblue:
```suggestion
`ember-test-waiters` uses a minimal API to provide waiting functionality. This minimal API can be composed to accommodate various complex scenarios.
```

README.md
```diff
@@ -1,37 +1,71 @@
...
-Usage
-------------------------------------------------------------------------------
+The `TestWaiter` class is, in most cases, all you will need to wait for async operations to complete before continuing tests.
 
-[Longer description of how to use the addon in apps.]
+```
+import Component from '@ember/component';
+import { TestWaiter } from 'ember-test-waiters';
+
+if (DEBUG) {
```
rwjblue:
```suggestion
let waiter = DEBUG && new TestWaiter('friend-waiter');
```

README.md
```diff
@@ -1,37 +1,71 @@
...
+  let waiter = new TestWaiter('friend-waiter');
```
rwjblue:
```suggestion
```

README.md
```diff
@@ -1,37 +1,71 @@
...
+}
```
rwjblue:
```suggestion
```

addon/test-waiter.ts
```diff
@@ -0,0 +1,116 @@
+import { IWaiter, WaiterName, ITestWaiterDebugInfo } from './types';
+import { register } from './waiter-manager';
+
+/**
+ * A class providing creation, registration and async waiting functionality.
```
rwjblue: I hate hate hate hate hate this style of comment 😩

addon/test-waiter.ts
```diff
@@ -0,0 +1,116 @@
...
+ *
+ * @public
+ * @class SimpleWaiter<T>
+ *
+ * @example
+ *
+ * import Component from '@ember/component';
```
rwjblue: Specifically, these leading asterisks make the examples hard to review and break the ability to copy paste.

addon/test-waiter.ts
```diff
@@ -0,0 +1,116 @@
...
+ * import { SimpleWaiter } from 'ember-test-waiters';
+ *
+ * if (DEBUG) {
```
rwjblue: Same change as I suggested in the README, should be something like:

```js
let waiter = DEBUG && new SimpleWaiter('friend-waiter');
```

addon/test-waiter.ts
```diff
@@ -0,0 +1,116 @@
...
+ *   let waiter = new SimpleWaiter('friend-waiter');
+ * }
+ *
+ * export default class Friendz extends Component {
+ *   didInsertElement() {
+ *     waiter.beginAsync(this);
```
rwjblue: Usages of `waiter` should also be guarded by `DEBUG`:

```suggestion
       DEBUG && waiter.beginAsync(this);
```

addon/test-waiter.ts
```diff
@@ -0,0 +1,116 @@
...
+ *
+ *     someAsyncWork().then(() => {
+ *       waiter.endAsync(this);
```
rwjblue:
```suggestion
         DEBUG && waiter.endAsync(this);
```

addon/test-waiter.ts
```diff
@@ -0,0 +1,116 @@
...
+/**
+ * A class providing creation, registration and async waiting functionality.
+ *
+ * @public
+ * @class SimpleWaiter<T>
+ *
+ * @example
+ *
+ * import Component from '@ember/component';
+ * import { SimpleWaiter } from 'ember-test-waiters';
```
rwjblue: Is the export actually "SimpleWaiter"? Seems like its called `TestWaiter` below...

addon/test-waiter.ts
```diff
@@ -0,0 +1,116 @@
...
+  /**
+   * Should be used to signal the end of an async operation. Invocation of this
+   * method should be paired with a preceeding `beginAsync` call, which would indicate the
+   * beginning of an async operation.
+   *
+   * @public
+   * @method endAsync
+   * @param item {T} The item to that was registered for waiting
+   */
+  endAsync(item: T) {
```
rwjblue: Should we do anything if `endAsync` is called and the specific `item` is not found within our Map?

addon/wait-for-promise.ts
```diff
@@ -0,0 +1,45 @@
...
+
+/**
+ * Allows for tracking the beginning and end of async operations within promises.
+ *
+ * @public
+ * @param promise {Promise<T>} The promise to track async operations for
+ * @param label {string} An optional string to identify the promise
+ * @example
+ *
+ * import Component from '@ember/component';
```
rwjblue: :rage4: `*`

addon/wait-for-promise.ts
```diff
@@ -0,0 +1,45 @@
+import { DEBUG } from '@glimmer/env';
+import TestWaiter from './test-waiter';
+
+const PROMISE_WAITER = new TestWaiter('promise-waiter');
+
+/**
+ * Allows for tracking the beginning and end of async operations within promises.
```
rwjblue: I don't think this is quite correct. How does this sound:

```suggestion
  A convenient utility function to simplify waiting for a promise.
```

addon/waiter-manager.ts
```diff
@@ -0,0 +1,79 @@
+import { IWaiter, IPendingWaiterState } from './types';
+
+const WAITERS = new Set();
+
+/**
+ * Registers a waiter with the test waiter manager.
```
rwjblue:
```suggestion
 * Registers a waiter.
```

addon/waiter-manager.ts
```diff
@@ -0,0 +1,79 @@
...
+ *
+ * @public
+ * @param waiter {IWaiter} A test waiter instance
+ */
+export function register(waiter: IWaiter): void {
+  WAITERS.add(waiter);
+}
+
+/**
+ * Unregisters a waiter with the test waiter manager.
```
rwjblue:
```suggestion
 * Unregisters a waiter.
```

addon/waiter-manager.ts
```diff
@@ -0,0 +1,79 @@
...
+ *
+ * @public
+ * @param waiter {IWaiter} A test waiter instance
+ */
+export function unregister(waiter: IWaiter): void {
+  WAITERS.delete(waiter);
+}
+
+/**
+ * Gets an array of all test waiters current registered.
```
rwjblue:
```suggestion
 * Gets an array of all waiters current registered.
```

addon/waiter-manager.ts
```diff
@@ -0,0 +1,79 @@
...
+ *
+ * @public
+ * @returns {IWaiter[]}
+ */
+export function getWaiters(): IWaiter[] {
+  return [...WAITERS];
+}
+
+/**
+ * Clears all test waiters from the waiter manager.
```
rwjblue:
```suggestion
 * Clears all waiters.
```

addon/waiter-manager.ts
```diff
@@ -0,0 +1,79 @@
...
+ *
+ * @public
+ */
+export function reset(): void {
+  WAITERS.clear();
+}
+
+/**
+ * Gets the current state of all test waiters. Any test waiters whose
```
rwjblue:
```suggestion
 * Gets the current state of all waiters. Any waiters whose
```

addon/waiter-manager.ts
```diff
@@ -0,0 +1,79 @@
...
+ * `waitUntil` method returns false will be considered `pending`.
+ *
+ * @returns a result object containing a count of all waiters pending and
```
rwjblue: Is there a better way to type this (for docs I mean)? Can we give the return type an interface, and reference it here?

addon/waiter-manager.ts
```diff
@@ -0,0 +1,79 @@
...
+      let debugInfo = waiter.debugInfo();
+      result.waiters[waiter.name] = debugInfo || true;
+    }
+  });
+
+  return result;
+}
+
+/**
+ * Determines if there are any pending waiters in the waiter manager.
```
rwjblue:
```suggestion
 * Determines if there are any pending waiters.
```

addon/index.ts
```diff
@@ -4,8 +4,8 @@ export {
   getWaiters,
   reset,
```
rwjblue: Should probably be underscored, right?

tests/unit/waiter-manager-test.ts
```diff
@@ -0,0 +1,129 @@
...
+  unregister,
+  hasPendingWaiters,
+  getWaiters,
+  reset,
+  TestWaiter,
+  getPendingWaiterState,
+} from 'ember-test-waiters';
+import MockStableError, { overrideError, resetError } from './utils/mock-stable-error';
+
+module('test-watiers', function(hooks) {
```
rwjblue:
```suggestion
module('test-waiters', function(hooks) {
```

API.md
scalvert: Oh yes, absolutely. I didn't know how to filter that.

README.md
scalvert: Sure, but to be clear, this is taken verbatim from here:

https://github.com/emberjs/ember.js/blob/master/packages/ember-testing/lib/test/waiters.js#L8

addon/index.ts
scalvert: Ya, I wanted to go through the process of using these things before I determined, once and for all, what was private vs. public. I think we're close to determining this now.

API.md
scalvert: Fixed!

addon/waiter-manager.ts
scalvert: Done.

addon/index.ts
scalvert: Done.

README.md
scalvert: Updated. Let me know what you think.

scalvert:
I've left the prod build config commented out in the travis.yml until we can figure out the root case of its failure. Other than that I think we should be OK to merge this as is. Once it's in we can circle back on any small bits we want to change.

addon/test-waiter.ts
scalvert: It's the default when I add a jsdoc in vscode. I love love love love it!

addon/test-waiter.ts
scalvert: We should probably ensure that the `beginAsync` and `endAsync` calls are balanced. The question is, what do we do? I would think throwing would seem a sensible result. What do you think?

addon/wait-for-promise.ts
scalvert: ❄️ *

rwjblue:
> until we can figure out the root case of its failure 

sounds good, mind making an issue to keep track of that?

.travis.yml
```diff
@@ -45,6 +45,11 @@ jobs:
       script:
         - yarn test
 
+    #- stage: 'Production Tests'
```
rwjblue: Can you make an issue and make a `# TODO: Fix #1234` comment here?

addon/build-waiter.ts
```diff
@@ -0,0 +1,37 @@
...
+import { TestWaiter } from 'ember-test-waiters';
+import NoopTestWaiter from './noop-test-waiter';
+
+/**
+ * Builds and returns a test waiter. The type of the
+ * returned waiter is dependent on whether the app or
+ * addon is in `DEBUG` mode or not.
+ *
+ * @param name {string} The name of the test waiter
+ * @returns {TestWaiter|NoopTestWaiter}
```
rwjblue: I'd rather return an interface (instead of `SomeClass|OtherClass`), since both classes implement the same interface (`beginAsync` / `endAsync` / `register` / etc).

addon/build-waiter.ts
```diff
@@ -0,0 +1,37 @@
...
+ *   didInsertElement() {
+ *     waiter.beginAsync(this);
+ *
+ *     someAsyncWork().then(() => {
+ *       waiter.endAsync(this);
+ *     });
+ *   }
+ * }
+ */
+export default function buildWaiter(name: string) {
```
rwjblue: Can you add an explicit return type here? I know typescript can infer it, but when reading its nice to see...

addon/test-waiter.ts
rwjblue: I'll fix them all when you are done ;)

addon/test-waiter.ts
```diff
@@ -0,0 +1,97 @@
...
+   * @public
+   * @method beginAsync
+   * @param item {T} The item to register for waiting
+   * @param label {string} An optional label to identify the item
+   */
+  beginAsync(item: T, label?: string) {
+    this.register();
+
+    this.items.set(item, {
+      stack: new Error().stack,
```
rwjblue: Nothing to do here ATM, but I wonder if we should create the error and not forcefully/eagerly do the stack generation. In most cases we won't use the stack, and I believe generating one is somewhat costly...

addon/test-waiter.ts
rwjblue: Ya, lets throw (or assert) for now, if we find a use case that we hadn't accounted for we can always tweak.

addon/noop-test-waiter.ts
```diff
@@ -0,0 +1,19 @@
+import { IWaiter } from './types';
+
+export default class NoopTestWaiter implements IWaiter {
```
rwjblue: Can you create a specific interface that both `TestWaiter` and `NoopTestWaiter` implement? This way TS can verify things for us (that we are properly implementing all the expected methods and whatnot).

addon/test-waiter.ts
```diff
@@ -0,0 +1,97 @@
+import { IWaiter, WaiterName, ITestWaiterDebugInfo } from './types';
+import { register } from './waiter-manager';
+
+/**
+ * A class providing creation, registration and async waiting functionality.
+ *
+ * @public
+ * @class TestWaiter<T>
+ */
+export default class TestWaiter<T> implements IWaiter {
```
rwjblue: Same as above, can you make an interface that both this and noop waiter implement (that is more specific than `IWaiter`)?

addon/types/index.ts
```diff
@@ -3,10 +3,10 @@ export type WaiterName = string;
 export interface IWaiter {
```
rwjblue: Since we don't have a class for this, maybe we can just call this one `Waiter`

addon/build-waiter.ts
scalvert: Seems obvious in retrospect. For sure, will change.

addon/build-waiter.ts
scalvert: 👍

addon/noop-test-waiter.ts
scalvert: Done. I like this.

addon/test-waiter.ts
scalvert: Done.

addon/types/index.ts
scalvert: I'd prefer to keep our interface naming consistent, if you don't mind.

addon/test-waiter.ts
scalvert: Yep, this is what we do in Backburner, so I think we do the same here.

scalvert: Done.

.travis.yml
scalvert: There already was one. https://github.com/ember-cli/ember-cli/issues/8154. Added a comment to this file linking the issue.

addon/test-waiter.ts
scalvert: :)

addon/test-waiter.ts
scalvert: Done.

scalvert:
@rwjblue I think this is ready to go.

addon/types/index.ts
```diff
@@ -3,10 +3,15 @@ export type WaiterName = string;
...
   name: WaiterName;
   waitUntil(): boolean;
-  debugInfo(): unknown[] | undefined | void;
+  debugInfo(): unknown;
```
rwjblue: Hmm, don't we want to ensure it always returns an array (which would mean `unknown[]`)?

When this is used by @ember/test-helper's I'd rather be able to make the safe assumption that it always returns an array...

scalvert: If people define their own `IWaiter`, they may want to return their own specific type or format from `debugInfo`. If we require them to return an array, that's constraining them to a specific shape of data. I'm not sure we want to do that.

rwjblue: I'm unconvinced. Lets keep an eye on the work that this places on the `@ember/test-helpers` integration and see if we want to revisit...

PR #8, opened 2019-05-10
Converts the system to a token based system rather than arbitrary objects

scalvert:
This PR does two things:

1. Converts the `TestWaiter` class to a token based system to maintain uniqueness between calls (`beginAsync` and `endAsync` calls can be paired, rather than the same object value being passed to all calls to waiters, which resulted in errors thrown if multiple `endAsync` calls happened in a row)
1. Adds more tests for the token system


tests/test-helper.js
```diff
@@ -1,8 +1,29 @@
-import Application from '../app';
-import config from '../config/environment';
-import { setApplication } from '@ember/test-helpers';
-import { start } from 'ember-qunit';
+import QUnit from 'qunit';
+import AbstractTestLoader from 'ember-cli-test-loader/test-support/index';
```
step2yeung: im curious why we are using the base testloader here instead of using ember-qunit

README.md
```diff
@@ -29,7 +29,10 @@ ember install ember-test-waiters
 
 ### buildWaiter function
 
-The `buildWaiter` function is, in most cases, all you will need to wait for async operations to complete before continuing tests.
+The `buildWaiter` function is, in most cases, all you will need to wait for async operations to complete before continuing tests. It returns a waiter instance
```
step2yeung: `in most cases` are there cases u have in mind where this won't cover? should it be called out?

scalvert: Well yes, it would be any case that you would want to define your own custom waiter class. I believe that 99 times out of 100 you’ll just want to use buildWaiter.

README.md
```diff
@@ -39,18 +42,97 @@ let waiter = buildWaiter('friend-waiter');
...
+        })
+        .finally(() => {
+          waiter.endAsync(this);
+        });
+    }
+  }
+}
+```
+
+The following code _will_ guarantee uniqueness, as the two `endAsync` calls _may_ be called in succession, but have unique tokens:
```
rwjblue: I think we need to rephrase this stuff about "uniqueness", or at least explain what we mean before saying that the consumers should care about it. 

Frankly, there are absolutely times that you **do not** want uniqueness and instead you want to leverage our waiter system's assertion when two async things' ordering is not enforced.

addon/types/index.ts
```diff
@@ -1,14 +1,15 @@
...
 export interface IWaiter {
   name: WaiterName;
   waitUntil(): boolean;
   debugInfo(): unknown;
 }
 
-export interface ITestWaiter<T> extends IWaiter {
-  beginAsync(item: T, label?: string): void;
-  endAsync(item: T): void;
+export interface ITestWaiter extends IWaiter {
```
rwjblue: Why remove the generic? It still seems accurate, and removing it means that you _can't_ enforce types for the token if you want to...

addon/waiter-manager.ts
```diff
@@ -32,6 +32,16 @@ export function getWaiters(): IWaiter[] {
...
 }
 
+/**
+ * Puts waiters back into a default state where
+ * no waiter is pending.
+ *
+ * @public
+ */
+export function resetPendingWaiters(): void {
+  WAITERS.forEach(waiter => waiter.reset());
```
rwjblue: This isn't a valid operation, there is no `reset` method on either the `ITestWaiter` or `IWaiter` interfaces. The type for `WAITERS` is not setup correctly we need to change from:

```js
const WAITERS = new Map();
```

To:

```js
const WAITERS = new Map<string, IWaiter>();
```

tests/index.html
```diff
@@ -21,6 +21,13 @@
     {{content-for "body"}}
     {{content-for "test-body"}}
 
+    <div id="qunit"></div>
+    <div id="qunit-fixture"></div>
+
+    <div id="ember-testing-container">
```
rwjblue: This may not needed, what happens if you remove it?

tests/unit/test-waiter-test.ts
```diff
@@ -1,22 +1,46 @@
+// @ts-ignore
```
rwjblue: Why do we need this ts-ignore? Doesn't this make it ignore the whole file? 

Is this perhaps because the `qunit` module is unknown to the system? If so, we should add a `types/qunit.d.ts` file instead of ts-ignoring...

tests/unit/test-waiter-test.ts
```diff
@@ -1,22 +1,46 @@
...
 
-module('test-waiter', function(hooks) {
+module('test-waiter', function(hooks: NestedHooks) {
   hooks.afterEach(function() {
     _reset();
     resetError();
   });
 
-  test('test waiter can be instantiated with a name', function(assert) {
+  test('test waiter can be instantiated with a name', function(assert: Assert) {
```
rwjblue: Uhh, why do you have to type this? Seems related to the ts-ignore at the top, we should absolutely fix...

tests/unit/test-waiter-test.ts
```diff
@@ -1,22 +1,46 @@
...
     let waiter = new TestWaiter(name);
 
     assert.equal(waiter.name, name);
   });
 
-  test('test waiters automatically register when beginAsync is invoked', function(assert) {
+  test('test waiters return a token from beginAsync when no token provided', function(assert: Assert) {
+    let waiter = new TestWaiter('my-waiter');
+
+    let token = <number>waiter.beginAsync();
```
rwjblue: Why cast this to a number? I think we should change this around to assert the value is non-null and non-undefined (we don't care if its a number or not).

We also need some tests (or assertions in this test) to ensure `waiter.beginAsync() !== waiter.beginAsync()`.

tests/unit/wait-for-promise-test.ts
```diff
@@ -1,17 +1,18 @@
+// @ts-ignore
```
rwjblue: Same comment as above, we should ensure that the `qunit` module is properly typed

tests/unit/waiter-manager-noop-test.ts
```diff
@@ -1,21 +1,22 @@
+// @ts-ignore
```
rwjblue: 🔨

tests/unit/waiter-manager-noop-test.ts
```diff
@@ -24,14 +25,13 @@ if (!DEBUG) {
...
     });
 
-    test('a NoopTestWaiter always returns true from waitUntil', function(assert) {
+    test('a NoopTestWaiter always returns true from waitUntil', function(assert: Assert) {
       let waiter = buildWaiter('first');
-      let waiterItem = {};
 
       assert.ok(waiter.waitUntil(), 'waitUntil returns true');
-      waiter.beginAsync(waiterItem);
+      let token = waiter.beginAsync();
```
rwjblue: Instead of changing the meaning of this test, we should add another test (IOW, we should test both custom token and generated token versions).

tests/unit/waiter-manager-test.ts
```diff
@@ -1,3 +1,4 @@
+// @ts-ignore
```
rwjblue: 🔪

scalvert:
@rwjblue you're right, this PR got absurdly large. I'll pull out the different bits into their own PRs.

README.md
```diff
@@ -39,18 +42,97 @@ let waiter = buildWaiter('friend-waiter');
...
+        waiter.endAsync(token);
+      });
+  }
+}
+```
+
+#### Custom Tokens
+
+The `beginAsync` method also can receive a custom token, if you need to specifically define a token to represent your async operation. _Note_ - due to the nature of
+async operations, it's impossible to guarantee which order the `endAsync` calls will run. If you are making multiple `beginAsync`/`endAsync` calls wiithin a single
```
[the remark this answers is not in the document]
scalvert: What I mean to say here is it's not possible to ensure that these calls will not interleave, meaning that two separate pairs of `beginAsync`/`endAsync` won't interleave.

README.md
scalvert: What are times when you'd not want uniqueness, for a discrete async operation?

addon/types/index.ts
scalvert: You're right, `T` should be the generic type for token.

addon/waiter-manager.ts
scalvert: This was simply an omission on my part. I'll add.

tests/index.html
scalvert: Unsure. Will try.

tests/unit/test-waiter-test.ts
scalvert: Yes, this was an issue with `@types/qunit`, in that it hadn't declared QUnit as a module. That's been fixed upstream in DefinitelyTyped/DefinitelyTyped/pull/35350. 

I can revisit adding the module, but was hoping to just pull in the fix from upstream.

tests/unit/test-waiter-test.ts
scalvert: Correct. Hopefully it's resolved once the fix for `@types/qunit` is resolved.

tests/unit/test-waiter-test.ts
scalvert: Seems reasonable. Good suggestions.

tests/unit/wait-for-promise-test.ts
scalvert: Same as my answer above.

tests/unit/waiter-manager-noop-test.ts
scalvert: 🛑

tests/unit/waiter-manager-noop-test.ts
scalvert: 👍

tests/unit/waiter-manager-test.ts
scalvert: 🛑

README.md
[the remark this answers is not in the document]
rwjblue: Right, but the caller of `beginAsync` and `endAsync` **absolutely** can ensure they won't interleave. It may be hard, but it is absolutely possible...

tests/unit/test-waiter-test.ts
rwjblue: Ya, totally if they fixed it that seems great!

README.md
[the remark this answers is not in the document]
scalvert: I think I'm focusing on the common case here, where the average caller may not be thinking about the full breadth of their async operations (and yes, they should be!). I don't disagree with you, I just believe that the token system makes them fall into a pit of success in a much easier way.

rwjblue:
Landed the dropping of ember-qunit over in #9

tests/unit/test-waiter-test.ts
mike-north: I need to get around to a fix for [the `ember-qunit` types](https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/types/ember-qunit). @rwjblue pointed out that any "ES6 module-ification" of QUnit belongs there

tests/unit/test-waiter-test.ts
mike-north: this is a good place for a type guard
```ts
if (typeof token !== 'number') {
   throw new Error('Non-numeric token: ' + token);
}
// token will be of type number here

```

tests/unit/wait-for-promise-test.ts
mike-north: > Same comment as above, we should ensure that the qunit module is properly typed

I hope get this done thurs/fri this week at latest

README.md
mike-north:
```suggestion
async operations, it's impossible to guarantee which order the `endAsync` calls will run. If you are making multiple `beginAsync`/`endAsync` calls within a single
```

addon/types/index.ts
mike-north: @scalvert and I discussed this, and I feel pretty strongly that `unknown` is appropriate for the type of the token, given that we allow consumers of the library to hold onto it, but don't want them to depend on its specific value.

Enforcing types for the token is possible _through type guards_, and which is appropriate in my opinion b/c it's only meaningful at runtime.

addon/waiter-manager.ts
scalvert: I can update in a separate PR.

tests/index.html
scalvert: We pulled these changes out into a separate PR - #9

tests/unit/test-waiter-test.ts
scalvert: This has been cleaned up as part of the test PR, #9

tests/unit/wait-for-promise-test.ts
scalvert: We managed to figure out a stop-gap solution as part of #9

tests/index.html
scalvert: Going to push this to a later PR.

tests/unit/test-waiter-test.ts
scalvert: We've removed the need to cast now.

addon/types/index.ts
scalvert: We've managed to come up with a happy medium, where we default `T` to `Token`.

README.md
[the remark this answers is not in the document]
scalvert: Updated the prose.

README.md
scalvert: Removed custom waiter info from the README, for now.

ember-cli-build.js
```diff
@@ -3,17 +3,16 @@
 const EmberAddon = require('ember-cli/lib/broccoli/ember-addon');
 
 module.exports = function(defaults) {
-  let app = new EmberAddon(defaults, {});
-
-  app.import('node_modules/qunit/qunit/qunit.js', {
```
rwjblue: This seems wrong 🤔

package.json
```diff
@@ -45,13 +45,13 @@
...
     "ember-cli-inject-live-reload": "^1.8.2",
     "ember-cli-sri": "^2.1.1",
-    "ember-cli-test-loader": "^2.2.0",
     "ember-cli-typescript": "^1.5.0",
     "ember-cli-uglify": "^2.1.0",
     "ember-disable-prototype-extensions": "^1.1.3",
     "ember-export-application-global": "^2.0.0",
     "ember-load-initializers": "^1.1.0",
     "ember-maybe-import-regenerator": "^0.1.6",
+    "ember-qunit": "^3.4.1",
```
rwjblue: Seems like a bad rebase?

package.json
```diff
@@ -63,7 +63,7 @@
     "eslint-plugin-prettier": "^3.0.1",
     "loader.js": "^4.7.0",
     "prettier": "^1.16.4",
-    "qunit": "^2.9.2",
+    "qunit-dom": "^0.8.0",
```
rwjblue: Didn't you remove this?

addon/test-waiter.ts
```diff
@@ -1,25 +1,34 @@
...
  * @class TestWaiter<T>
  */
-export default class TestWaiter<T> implements ITestWaiter<T> {
+export default class TestWaiter<T = Token> implements ITestWaiter<T> {
   public name: WaiterName;
+  private nextToken: () => T;
   private isRegistered = false;
 
-  items = new Map<T, ITestWaiterDebugInfo>();
+  items = new Map<Token, ITestWaiterDebugInfo>();
```
rwjblue: Should be `T` 
```suggestion
  items = new Map<T, ITestWaiterDebugInfo>();
```

PR #11, opened 2019-05-24
Refining types for waiter-manager

scalvert:
The types in `waiter-manager.ts` were a little loose. This PR tightens them up a bit.

### DIFF
--- a/addon/types/index.ts
+++ b/addon/types/index.ts
@@ -4,7 +4,7 @@ export type Token = unknown;
 export interface IWaiter {
   name: WaiterName;
   waitUntil(): boolean;
-  debugInfo(): unknown;
+  debugInfo(): ITestWaiterDebugInfo[];
 }
 
 export interface ITestWaiter<T = Token> extends IWaiter {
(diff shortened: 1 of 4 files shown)


addon/types/index.ts
```diff
@@ -4,7 +4,7 @@ export type Token = unknown;
 export interface IWaiter {
   name: WaiterName;
   waitUntil(): boolean;
-  debugInfo(): unknown;
+  debugInfo(): ITestWaiterDebugInfo[];
```
rwjblue: I'd rather the generic thing not be forced into the same structure as what TestWaiter would use.

scalvert: I thought about this too, but from a consumer standpoint, we expect to output this debug info as part of `@ember/test-helpers`. Specifically, _that_ library _does_ expect this to be a particular format. In fact, it exactly expects this format, as demonstrated [here](https://github.com/emberjs/ember-test-helpers/blob/master/addon-test-support/%40ember/test-helpers/-internal/debug-info.ts#L157). How would we handle this otherwise?

rwjblue: Kk, seems good then. Thank you!

PR #13, opened 2019-08-02
task(compatibility): Enabling compatibility with versions that are pre-ember-test-waiters

scalvert:
Allows for compatibility for apps and addons that are pre-`@ember/test-helpers` 1.6.0, which was the first version that included `ember-test-waiters`.

Closes https://github.com/rwjblue/ember-test-waiters/issues/12


scalvert:
It appears CI is failing due to `ember-compatibility-helpers` expecting `@ember/test-helpers` to be either a `dependency` or a `devDependency`. We intentionally avoided that with this addon, so I'm not sure the best way to address this. Thoughts?

addon/waiter-manager.ts
```diff
@@ -1,7 +1,17 @@
...
 const WAITERS: Map<WaiterName, IWaiter> = new Map<WaiterName, IWaiter>();
 
+/**
+ * Backwards compatibility with legacy waiters system.
+ *
+ * We want to always register a waiter using the legacy waiter system, as right
+ * now if consumers are not on the right version of @ember/test-helpers, using
+ * this addon will result in none of these waiters waiting.
+ */
+registerWaiter(hasPendingWaiters);
```
rwjblue: I actually think we need to guard this a tad:

```suggestion
if (Ember.Test) {
  registerWaiter(hasPendingWaiters);
}
```

Specifically we should use `if (Ember.Test)` (likely with eslint ignore shenanigans) so that we do not error in production builds (where invoking `registerWaiter` unconditionally like this will actually error) and we would prefer to avoid using `DEBUG` because of `ember test -prod`.

testem.js
```diff
@@ -9,7 +9,6 @@ module.exports = {
         // --no-sandbox is needed when running Chrome inside a container
         process.env.CI ? '--no-sandbox' : null,
         '--headless',
-        '--disable-gpu',
```
rwjblue: :sob:

addon/waiter-manager.ts
rwjblue: I created https://github.com/rwjblue/ember-test-waiters/issues/14 to track ensuring we fail PR's that would break production non-testing builds.

scalvert: Ah yes. Good call.

testem.js
scalvert: I know. Chrome 76 goes 💥

PR #17, opened 2019-08-03
Reverse legacy waiter check

### LINKED ISSUE #16: CI test timeout
Hi,

CI for `ember-web-app` is failing now, because test timeout https://travis-ci.org/zonkyio/ember-web-app/jobs/567242322.

The [registration](https://github.com/rwjblue/ember-test-waiters/blob/master/addon/waiter-manager.ts#L16) of legacy test waiter is causing tests to timeout, because it expects `true` if there are no pending tasks, but [hasPendingWaiters()](https://github.com/rwjblue/ember-test-waiters/blob/master/addon/waiter-manager.ts#L92) returns `true`, when it has some pending tasks to wait for.

bobisjan:
fixes #16

### DIFF
--- a/addon/waiter-manager.ts
+++ b/addon/waiter-manager.ts
@@ -13,7 +13,7 @@ const WAITERS: Map<WaiterName, IWaiter> = new Map<WaiterName, IWaiter>();
  */
 // eslint-disable-next-line ember/new-module-imports
 if (Ember.Test) {
-  registerWaiter(hasPendingWaiters);
+  registerWaiter(() => !hasPendingWaiters());
 }
 
 /**


bobisjan:
I'm not sure about a test for this, because project is not using `ember-qunit`?

Based on `ember-web-app` tests, some sanity acceptance test should cover this case:

```js
test('sanity', async function(assert) {
  await visit('/');
  assert.equal(currentURL(), '/');
});
```

scalvert:
Oh wow, thanks for this! The irony is that this is precisely the reason @rwjblue and I created this update to the waiters API in the first place; the API was such that it was really easy to make the mistake of reversing your logic when waiting. 

I really appreciate you adding this. 

As for tests, I'll just merge this in for now and we can figure out the testing strategy later. We intentionally don't want to rely on `ember-qunit` or `@ember/test-helpers`, as this library is intended to stand alone from those.

rwjblue:
:sob:

bobisjan:
@scalvert thanks for a quick release 👍.

PR #69, opened 2019-10-22
Do not throw errors for endAsync being called multiple times

### LINKED ISSUE #4: Determine whether we need to track when `endAsync` is called more than once.
We currently only error if `endAsync` is called without calling a `beginAsync` in the `TestWaiter` class. It may be the case that we find we need more fine grained error messaging for two scenarios:

1. `endAsync` is called before `beginAsync` is called, thus not registering the waiter item (implemented)
2. `endAsync` is called more than once for the same waiter item (not implemented).

seanjohnson08:
Fixes #4

### DIFF
--- a/package.json
+++ b/package.json
@@ -68,7 +68,7 @@
     "qunit": "^2.9.2",
     "release-it": "^12.2.1",
     "release-it-lerna-changelog": "^1.0.3",
-    "typescript": "^3.5.3"
+    "typescript": "^3.7.4"
   },
   "resolutions": {
     "**/engine.io": "~3.3.0"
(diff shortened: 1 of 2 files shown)


addon/test-waiter.ts
```diff
@@ -19,6 +19,7 @@ export default class TestWaiter<T = Token> implements ITestWaiter<T> {
   private isRegistered = false;
 
   items = new Map<T, ITestWaiterDebugInfo>();
+  endedItems = new Map<T, boolean>();
```
rwjblue: Can we avoid needing the extra map here, by weakly associating the `TestWaiter` with the token?

scalvert: To be clear, Rob is suggesting we do

```ts
let trackedTokens = new WeakMap<Token, TestWaiter>();
```

to track the association of the token with a particular test waiter instance, _and_ have the ability to detect if a token is attempting to be used to invoke `endAsync` more than once.

rwjblue:
@scalvert / @seanjohnson08 - Where are we at with this one?

scalvert:
@seanjohnson08?

seanjohnson08:
I'm closing this - I don't believe it's an issue for our use case any longer.

PR #81, opened 2019-11-28
Wait for coroutine

bendemboski:
- Implement `waitForCoroutine` that can be used in decorator or function-wrapper form
- Extend `waitForPromise` to be usable as a decorator wrapping an async member function
- Strip both out of builds that don't have tests using a babel transform


bendemboski:
I'm not sure why `waitForPromise`'s examples section is no longer in the docs, or why `waitForCoroutine` doesn't have one -- any ideas? Do the docs not support multiple `@example` sections or something?

bendemboski:
Okay, @rwjblue this is all passing tests now, and I think has everything we discussed. Still unsure about the documentation funniness -- any ideas? If not I can fiddle around to see if I can get it to do what we want, but hoping you might just know the answer 🤞

scalvert:
Thanks for this PR, @bendemboski! I think it's a nice addition/enhancement. 

A couple of things before I really look into the details of the changes
1. This PR contains some updates to ember-try configs, etc. I'd prefer those were done in a separate PR
2. I understand the necessity for the added complexity to support the decorator variants, but would like to see if we can figure out a way to reduce the added complexity in the utils like `waitForPromise`

Stay tuned, I'm still reviewing.

addon/wait-for-coroutine.ts
```diff
@@ -0,0 +1,74 @@
...
+  _key: string,
+  descriptor: PropertyDescriptor,
+  label?: string
+): PropertyDescriptor;
+export default function waitForCoroutine(fn: Function, label?: string): Function;
+
+export default function waitForCoroutine(
+  ...args: [object, string, PropertyDescriptor, string?] | [Function, string?]
+): PropertyDescriptor | Function {
+  if (args.length < 3) {
```
scalvert: I'd prefer if we stored this result in a variable that denoted that this was triggering the functional invocation branch of this code.

```ts
let isFunctionalInvocation = args.length < 3;

if (isFunctionalInvocation) {
  // ...
} else {
  // ...
}
```

addon/wait-for-coroutine.ts
```diff
@@ -0,0 +1,74 @@
...
+    // function
```
scalvert: If you do the above, you can delete this comment.

addon/wait-for-coroutine.ts
```diff
@@ -0,0 +1,74 @@
...
+    let [fn, label] = args as [Function, string?];
+
+    return wrap(fn, label);
+  } else {
+    // decorator
```
scalvert: And this comment.

addon/wait-for-promise.ts
```diff
@@ -1,10 +1,11 @@
...
 const PROMISE_WAITER = new TestWaiter<Promise<unknown>>('promise-waiter');
 
 /**
- * A convenient utility function to simplify waiting for a promise.
+ * A convenient utility function to simplify waiting for a promise. Can be used
+ * in both decorator and function form, and the function form either accepts a
+ * promise or an async function.
  *
  * @public
  * @param promise {Promise<T>} The promise to track async operations for
```
scalvert: Can we update the jsdoc to reflect the new signature?

addon/wait-for-coroutine.ts
```diff
@@ -0,0 +1,74 @@
...
+
+const COROUTINE_WAITER = new TestWaiter<Generator>('coroutine-waiter');
+
+/**
+ * A convenient utility function to simplify waiting for a coroutine in a
+ * context such as ember-concurrency. Can be used in both decorator and
+ * function form.
+ *
+ * @public
+ * @param coroutine {Generator<T, TReturn, TNext>} The coroutine to track async operations for
```
scalvert: Can we make the jsdoc params match the signature? Specifically, the `coroutine` param is actually `target` below, no?

addon/wait-for-promise.ts
```diff
@@ -23,24 +24,83 @@ const PROMISE_WAITER = new TestWaiter<Promise<unknown>>('promise-waiter');
...
+  label?: string
+): PropertyDescriptor;
+export default function waitForPromise<T>(promise: Promise<T>, label?: string): Promise<T>;
+export default function waitForPromise<T>(fn: Function, label?: string): Function;
+
+export default function waitForPromise<T>(
+  ...args:
+    | [object, string, PropertyDescriptor, string?]
+    | [Promise<T>, string?]
+    | [Function, string?]
```
scalvert: Should we define a public type for these tuples?

addon/wait-for-promise.ts
```diff
@@ -23,24 +24,83 @@ const PROMISE_WAITER = new TestWaiter<Promise<unknown>>('promise-waiter');
...
+): PropertyDescriptor | Promise<T> | Function {
+  if (args.length < 3) {
```
scalvert: Similar to my comment above, can we extract this condition into a variable that makes it clear that this is the functional invocation branch of the code?

addon/waiter-manager.ts
```diff
@@ -52,6 +52,7 @@ export function getWaiters(): IWaiter[] {
  * @public
  */
 export function _reset(): void {
+  WAITERS.forEach(waiter => ((waiter as any).isRegistered = false));
```
scalvert: This would be good in a separate PR. I can see why you're doing this; to ensure all waiter references have their properties set to `false`. It seems like a good edition, but I'd just prefer it's added outside of this PR as it muddies the waters.

config/ember-try.js
```diff
@@ -2,85 +2,90 @@
...
-              'jquery-integration': true,
-            }),
-          },
-          npm: {
-            devDependencies: {
-              '@ember/jquery': '^0.5.1',
-              'ember-source': '~2.18.0',
-            },
-          },
+module.exports = async function() {
```
scalvert: Again, would prefer these changes were done separately from this PR.

node-tests/babel-plugin-strip-wrappers-test.js
```diff
@@ -0,0 +1,175 @@
+const { expect } = require('chai');
```
scalvert: Why do we need to use mocha/chai for the node tests? We already use qunit here, and I'd rather we don't use both in the same project. Can we just use qunit?

package.json
```diff
@@ -38,39 +40,45 @@
...
     "qunit": "^2.9.2",
+    "tmp": "^0.1.0",
     "typescript": "^3.5.3"
   },
   "resolutions": {
     "**/engine.io": "~3.3.0"
   },
   "engines": {
-    "node": "6.* || 8.* || >= 10.*"
+    "node": "8.* || >= 10.*"
```
scalvert: I'd prefer that we did this change separately, in a different PR. Bumping the engine version, which I know is necessary here, requires a major version bump. Node 8 is almost out of LTS, so we'll have to do another one then also. Perhaps we can figure out how to avoid 2 majors in 1 month.

tests/unit/wait-for-coroutine-test.ts
```diff
@@ -0,0 +1,113 @@
...
+      runTests();
+    });
+
+    module('function', function(hooks) {
+      hooks.beforeEach(function() {
+        thing = FunctionThing.create();
+      });
+
+      runTests();
+    });
```
scalvert: I'd opt to pass a module name and type in as configuration in a loop, similar to how we did in ember-lifeline https://github.com/ember-lifeline/ember-lifeline/blob/3e331f97fa6a537a71993a918dc10900137944ab/tests/unit/dom-event-listeners-test.js#L43-L52

I guess I find it slightly easier to read if we do that.

.travis.yml
```diff
@@ -3,7 +3,7 @@ language: node_js
 node_js:
   # we recommend testing addons with the same minimum supported node version as Ember CLI
   # so that your addon works for all apps
-  - '6'
+  - "8"
```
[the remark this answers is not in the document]
bendemboski: Bumping the engine actually isn't necessary.

I just verified that we can make these changes without updating `ember-cli`/`ember-source`/etc (including the blueprint update which is what causes the engine update) if we use `ember-concurrency@1.0.0` and `ember-concurrency-decorators@0.6.0` in the tests (since anything newer won't run on Node 6). I'd prefer to run against the latest semver-compatibly 1.0s of these, but given that the code doesn't actually integrate with `e-c` directly, it's just the most common expected usage of `waitForCoroutine`, this is a mild preference.

As far as the major version bump goes, I guess I'd challenge the assumption that we'd need to do another major version bump in the next month. I'm following `ember-cli`'s blueprints here, and `ember-cli` dropped Node 6 from the blueprint 7 months ago, but we haven't had to update this library's blueprint/Node support until now (and per my investigation above, we don't even *have* to do it now). Are you anticipating some specific feature of Node 10 or 12 that we'll want to use in the next month that will require dropping support for Node 8? If not, then isn't it reasonable to assume that we could wait until at least April when Node 10 end-of-lifes before having to do another major version bump?

You've helped me realize that the Ember update is a separate change from this one, although it allows a cleaner/safer testing story IMO. It's just not clear to me that we gain anything by waiting for drop support for Node 6.

But I'm game either way:
1. Break out the `ember-cli`/blueprint update into a separate PR that gets merged before this one
2. Drop the `ember-cli`/blueprint update and use older versions of the `e-c` libraries for testing.

addon/wait-for-promise.ts
```diff
@@ -1,10 +1,11 @@
-import { DEBUG } from '@glimmer/env';
 import TestWaiter from './test-waiter';
 
 const PROMISE_WAITER = new TestWaiter<Promise<unknown>>('promise-waiter');
 
 /**
- * A convenient utility function to simplify waiting for a promise.
+ * A convenient utility function to simplify waiting for a promise. Can be used
```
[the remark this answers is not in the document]
bendemboski: I'm happy to make this change, but would point out that this is a request to change text that was not introduced in this PR...still want me to change it?

addon/waiter-manager.ts
bendemboski: This is actually required to make the tests pass. The issue was that the waiters would never re-register themselves once this was called because of [this check](https://github.com/rwjblue/ember-test-waiters/blob/873f7d577c34c47f33601b628e59564451a7c53d/addon/test-waiter.ts#L46). Previously only [one test](https://github.com/rwjblue/ember-test-waiters/blob/873f7d577c34c47f33601b628e59564451a7c53d/tests/unit/wait-for-promise-test.ts#L23-L33) was verifying that the promise waiter was actually waiting (and the [other two](https://github.com/rwjblue/ember-test-waiters/blob/873f7d577c34c47f33601b628e59564451a7c53d/tests/unit/wait-for-promise-test.ts#L40-L62) weren't actually testing anything because the waiter wasn't get re-registered after the first test). My changes cause multiple tests to verify that the waiter is installed, and all but the first will fail without this change.

So I guess I can break this out into a separate PR that this one would depend on if you like, but that's the reason that it's in this PR.

package.json
bendemboski: FYI, many of the changes that you've requested go in a separate PR are all the result of `ember-cli-update` applying the latest addon blueprint, so they can all collapse into https://github.com/rwjblue/ember-test-waiters/pull/81#discussion_r353887193

tests/unit/wait-for-coroutine-test.ts
bendemboski: I could do that, but since they have different setup steps (`beforeEach()`), it would be a bit more convoluted than the `ember-lifeline` example, since it would involve passing in a function with the `beforeEach` logic. IMO this reads more cleanly, but I'm sure it's stylistic, and this isn't my project, so if you still prefer the other way I'm happy to switch.

addon/wait-for-coroutine.ts
bendemboski: 👍 makes sense, will do

tests/unit/wait-for-coroutine-test.ts
scalvert: This is definitely in the realm of 'overly pedantic' in terms of should we do this vs. the structure in ember-lifeline. I'm inclined to say just ignore this, as I can go either way. 

It struck me that it took me a minute to orient myself on how the tests were structured, which is the reason for my comment. This isn't a make or break thing for me, unlike the choice of using mocha/chai above which is :)

.travis.yml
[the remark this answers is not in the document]
scalvert: Ya so typically we try to move node versions in lockstep with current LTS versions. As of 12/31, node 8 will be EOL (https://nodejs.org/en/about/releases/). 

If this change requires dropping node 6, and I do believe it does as the floating dependency tests will fail without that change due to another library removing node 6 support but not major version bumping, we're going to have to do a major version bump in _this_ library. Once node 8 EOLs, we're going to have to do another major bump. 

Does that make sense?

addon/wait-for-promise.ts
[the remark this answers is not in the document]
scalvert: Fair. Ignore. :)

addon/wait-for-coroutine.ts
scalvert: Thank you!

addon/waiter-manager.ts
scalvert: Hmmm ok. This does seem like a bug, as the waiters state wouldn't be reset in all the references if we didn't do this. I think this is a good catch. 

The reason I keep asking to split this apart into separate PRs is this one is quite large, and does seem to span a few different types of changes - blueprint updates, node version changes, bugfixes, and feature additions. It's just a bit hard to parse the whole thing. 

I'll leave it up to you. I won't die if this stays in.

package.json
scalvert: Yep, it'd be awesome to do those maintenance type tasks separately, as they're adding a bit of extra clutter to this PR. Does this seem fair?

tests/unit/wait-for-coroutine-test.ts
bendemboski: 👍 Will definitely convert to QUnit once we've collected/processed the feedback and it's time for another coding pass. I chose mocha because I'm much more used to seeing Node tests written in mocha (although I generally prefer QUnit myself) and thought that would keep it more familiar to people accustomed to Node and babel plugin testing and whatnot, but watch how eagerly I ditch it and replace with QUnit 😃

addon/waiter-manager.ts
bendemboski: Heh, I really don't feel strongly here. I'll tell you my, probably over-pedantic, hesitation to break it out into a separate PR (warning, some insight into the sometimes-silly inner workings of my brain ahead).

I really don't like committing code without test coverage, but since this, as far as I can tell, only exists for test instrumentation, I hesitate to write a test to test the test instrumentation. But maybe that's the right way to go? Or maybe just PR it without test coverage?

When it's part of this PR, I feel better because there is actually a test that will fail if this code doesn't exist (even though it's not explicitly testing the instrumentation...).

So I end up with decision paralysis, and would by default just leave it in, but can be swayed by the tiniest nudge 😜

node-tests/babel-plugin-strip-wrappers-test.js
bendemboski: Yes!

.travis.yml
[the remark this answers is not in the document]
bendemboski: Okay. I'm not following why another major bump will be required when Node 8 EOLs given that Node 6 EOLed 7 months ago and we haven't major bumped yet...but no worries, I don't need to understand and definitely don't want to be [this guy](https://xkcd.com/386/)!

I'm 90% sure that this change doesn't require dropping Node 6. The floating dependencies builds have been passing up until now (in a shocking and unprecedented plot twist, semver appears to be working!), and I can change the only dependencies that I need to add (`e-c` and `e-c-decorators` as `devDependencies`) to be pinned to versions that support Node 6. So we can stay on Node 6 until Node 8 EOLs, `ember-cli` updates the blueprint to drop 8 from `engines`, and then we can `ember-cli-update` at that point and just skip right over Node 8.

I'll make a backup copy of this branch, and then rebase away the `ember-cli`/blueprint update, pin the versions, and turn this into a more narrow PR.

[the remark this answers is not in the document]
bendemboski: (BTW, I just ran `ember try:one ember-default` locally with the dependency pinning I mentioned, and it passed, so now 95% sure we can merge this and stay on Node 6).

addon/waiter-manager.ts
scalvert: Ya I'm actually with you on the 'not testing the testing impl'. Let's keep it here. Done!

.travis.yml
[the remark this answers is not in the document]
scalvert: We discussed this offline, and can fix this by simply updating the travis.yml to run against node 8. This is done in another PR, which should likely get merged first.

addon/wait-for-promise.ts
bendemboski: I'm actually very lost as far as the documentation goes typescript + jsdoc + overloaded function signatures = a very confused Ben, and googling hasn't really helped. I think I'm going to need someone that understands it all to help out here.

addon/wait-for-promise.ts
bendemboski: 🤷‍♂ I'm new to typescript, so I'm not sure the pros/cons, or exactly how I would do it...

scalvert:
@bendemboski would you mind resolving conflicts in this PR? It'd be awesome to get this merged!

bendemboski:
@scalvert done!

scalvert:
Thanks. I'll take another pass over this tomorrow. Still wish it was split apart a bit eg. decorator support, new APIs etc in separate PRs. Don't worry about it now, but would be great to do this in the future!

README.md
```diff
@@ -78,6 +78,89 @@ export default class MoreFriendz extends Component {
...
+    this.doAsyncStuff().then(() => {
+      doOtherThings();
+    });
+  }
+})
+```
+
+### waitForCoroutine function
+
+This addon also provides a `waitForCoroutine` function, which can be used to wrap a coroutine, such as an `ember-concurrency` task function, so calls to it are registered with the test waiter system.
```
rwjblue: Can you link out to a definition of "coroutine" here?

rwjblue: I think we also have to be a bit more specific about what it means to wait for the coroutine. 

Does it wait for completion? Does it wait for the first yield? Something else entirely?

addon/wait-for-coroutine.ts
```diff
@@ -0,0 +1,74 @@
+import TestWaiter from './test-waiter';
+
+const COROUTINE_WAITER = new TestWaiter<Generator>('coroutine-waiter');
+
+/**
+ * A convenient utility function to simplify waiting for a coroutine in a
+ * context such as ember-concurrency. Can be used in both decorator and
+ * function form.
```
rwjblue: Let's beef this up to be more clear about what waiting means in this context. I'd like to see a dumbed down "for example, in the coroutine below, the decorator will beginAsync on line X and endAsync on line Y".

addon/wait-for-promise.ts
scalvert: I think we can actually leave this. It does a good job identifying the specific arguments for the 3 call signatures. Let's leave it.

addon/wait-for-coroutine.ts
```diff
@@ -0,0 +1,74 @@
...
+  } else {
+    let [, , descriptor, label] = args as [object, string, PropertyDescriptor, string];
+    let fn = descriptor.value;
+    descriptor.value = wrap(fn, label);
+    return descriptor;
+  }
+}
+
+function wrap(fn: Function, label?: string) {
+  return function*(this: any, ...args: any[]) {
```
rwjblue: I'm curious how this works in IE11, have you tested? Does this mean that we **require** regenerator runtime?

babel-plugin-strip-wrappers.js
```diff
@@ -0,0 +1,57 @@
+const syntaxDecorators = require('@babel/plugin-syntax-decorators').default;
```
rwjblue: Where does this actually come from? I think we need to be a bit more specific here. This currently expects the "top level" version to be what we want, but I think in reality we want the one from ember-cli-babel itself, right?

.eslintrc.js
```diff
@@ -61,5 +63,31 @@ module.exports = {
...
+      files: ['node-tests/**/*.js'],
+      env: {
+        qunit: true,
+      },
+      rules: Object.assign({}, require('eslint-plugin-node').configs.recommended.rules, {
+        'node/no-unpublished-require': 'off',
+        'node/no-extraneous-require': [
+          'error',
+          {
+            allowModules: ['@babel/core', '@babel/plugin-syntax-decorators'],
```
rwjblue: I'm not keen on allowing these to be required directly without a dep on them. Any reason we can't require the babel plugin to pass them through by way of `require.resolve`/`resolve.sync`?

index.js
```diff
@@ -9,5 +11,30 @@ module.exports = {
     this._super.included.apply(this, arguments);
 
     forceHighlander(this.project);
+
+    let app = this._findHost();
+    if (!app.tests) {
+      let checker = new VersionChecker(this.parent).for('ember-cli-babel', 'npm');
```
rwjblue: I wonder if we even need `ember-cli-version-checker` here, I would have written this like:

```js
let babelAddon = this.parent.addons.find(a => a.name === 'ember-cli-babel');
let babelVersion = babelAddon && babelAddon.pkg.version;
if (babelVersion && semver.gte(babelVersion, '7.0.0')) {

}
```

Basically, ember-cli-version-checker is checking disk but we actually have access to the version in memory.

tests/unit/wait-for-coroutine-test.ts
```diff
@@ -0,0 +1,115 @@
+import { module, test } from 'qunit';
+import { _reset, waitForCoroutine, getPendingWaiterState } from 'ember-test-waiters';
+import MockStableError, { resetError, overrideError } from './utils/mock-stable-error';
+import { task } from 'ember-concurrency-decorators';
+import EmberObject, { get } from '@ember/object';
+import { task as taskFn } from 'ember-concurrency';
+import { DEBUG } from '@glimmer/env';
+
+if (DEBUG) {
```
rwjblue: Would you mind moving the guard _inside_ the module?

Something like:

```js
module('wait-for-coroutine', function(hooks) {
  if (!DEBUG) {
    return;
  }

  // ...existing code...
});
```

tests/unit/wait-for-promise-test.ts
```diff
@@ -2,6 +2,7 @@ import { module, test } from 'qunit';
 import { Promise } from 'rsvp';
 import { _reset, waitForPromise, getPendingWaiterState } from 'ember-test-waiters';
 import MockStableError, { resetError, overrideError } from './utils/mock-stable-error';
+import EmberObject from '@ember/object';
 import { DEBUG } from '@glimmer/env';
 
 if (DEBUG) {
```
rwjblue: Whoops, now I see where this pattern came from. I do prefer to do this inside the module call though.

package.json
```diff
@@ -66,12 +72,15 @@
     "loader.js": "^4.7.0",
     "prettier": "^1.18.2",
     "qunit": "^2.9.2",
+    "qunitjs": "^2.4.1",
```
rwjblue: We should use `qunit` not `qunitjs` (`qunitjs` is deprecated in favor of `qunit`).

node-tests/babel-plugin-strip-wrappers-test.js
```diff
@@ -0,0 +1,181 @@
+const QUnit = require('qunitjs');
+const babel = require('@babel/core');
+const StripWrappers = require('../babel-plugin-strip-wrappers');
+const tmp = require('tmp');
+const { spawnSync } = require('child_process');
```
rwjblue: Lets use `execa` instead of `child_process`.

addon/wait-for-coroutine.ts
scalvert: 👍

README.md
scalvert: I also had to look this up! I mean I _knew_ what a coroutine was in terms of the definition, but I had to read the words to overlay the mental picture onto what this code is doing. I think a more robust description would be ideal.

addon/wait-for-coroutine.ts
scalvert: We should likely broaden our tests to encompass `windows-latest`?

tests/unit/wait-for-promise-test.ts
scalvert: Ya I think I can take credit for doing it that way.

index.js
```diff
@@ -9,5 +11,30 @@ module.exports = {
...
+      app.options = app.options || {};
+      if (checker.isAbove('7.0.0')) {
+        app.options.babel = app.options.babel || {};
+        app.options.babel.plugins = app.options.babel.plugins || [];
+        app.options.babel.plugins.unshift([StripWrappers, { legacy: true }]);
+      }
+    }
+  },
+
+  _findHost() {
```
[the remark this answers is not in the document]
bendemboski: Because it's a private method, so and I didn't want to create a dependency on the interface to a private method, possibly causing a dependency on a specific minor/bugfix version of `ember-cli` in the future if that private method's signature is ever changed.

addon/wait-for-coroutine.ts
bendemboski: I've never really bottomed out my understanding of regenerator runtime, maybe-import-, etc., so you may have to help educate me a bit. You mean because `function*` might fail to even parse? Cause nobody should even be importing this module if they don't already have a coroutine to wrap, which implies they have generator support. But if you're saying even including the addon would cause a runtime error, I understand, but would need some guidance on what to do about it.

babel-plugin-strip-wrappers.js
bendemboski: Oh yeah, I guess it's pretty accidental:

```sh
~/projects/ember-test-waiters$ yarn why @babel/plugin-syntax-decorators
yarn why v1.21.1
[1/4] 🤔  Why do we have the module "@babel/plugin-syntax-decorators"...?
[2/4] 🚚  Initialising dependency graph...
[3/4] 🔍  Finding dependency...
[4/4] 🚡  Calculating file sizes...
=> Found "@babel/plugin-syntax-decorators@7.7.4"
info Reasons this module exists
   - "documentation#@babel#plugin-proposal-decorators" depends on it
   - Hoisted from "documentation#@babel#plugin-proposal-decorators#@babel#plugin-syntax-decorators"
info Disk size without dependencies: "20KB"
info Disk size with unique dependencies: "48KB"
info Disk size with transitive dependencies: "48KB"
info Number of shared dependencies: 1
✨  Done in 0.55s.
```

And `documentation` is a `devDependency`, so at runtime we'll be getting it from 🤷‍♂ 

Are you suggesting something like

```javascript
const emberCliBabelPath = require.resolve('ember-cli-babel');
const syntaxDecoratorsPluginPath = require.resolve('@babel/plugin-syntax-decorators', {
  paths: [ emberCliBabelPath ]
});
```

?

.eslintrc.js
bendemboski: Ah, we're talking about the same thing as in [this comment](https://github.com/rwjblue/ember-test-waiters/pull/81/files#r377864336), right?

index.js
bendemboski: 🤷‍♂ I'm a little out of my depth here. I modeled this after how [ember-test-selectors](https://github.com/simplabs/ember-test-selectors/blob/master/index.js#L55-L74) does it, but will do whatever you think is best.

tests/unit/wait-for-coroutine-test.ts
bendemboski: Sure. Curious -- why do you prefer that?

scalvert:
👋 👋 👋 @bendemboski 👋 👋 👋

bendemboski:
@scalvert it's easier for me to do work like this in larger batches rather than bit by bit, so I'm waiting for answers to questions I had on some of the feedback from the last round before addressing.

index.js
scalvert: Ya so I think what Rob is saying is that addon discovery already loads all the addon metadata as part of the build, so we don't have to use version checker here; we can instead just use the data already loaded in memory.

scalvert:
Would you mind rebasing these changes with master, @bendemboski?

bendemboski:
This has gotten stale enough that I think it's probably no more effort to break apart into multiple PRs than to lift this monolithic one to merge-ability (and IIRC breaking it apart will make @scalvert happy anyway!), so I've completed the first step, which is extracting the `waitForPromise()` improvements [here](https://github.com/emberjs/ember-test-waiters/pull/131).

bendemboski:
[here](https://github.com/emberjs/ember-test-waiters/pull/132) is the code to strip the wrappers from production builds.

scalvert:
@bendemboski those other PRs are all merged in if you want to take a stab at the generator addition to `waitFor`. Thanks!

PR #84, opened 2019-12-04
Add public API for clearing waiter items

scalvert:
There are certain instances where we want to force reset all waiter items in order to ensure that waiters are not incorrectly blocking settledness in tests. This PR adds the `reset` method to the `TestWaiter` class to allow for resetting.

### DIFF
--- a/addon/noop-test-waiter.ts
+++ b/addon/noop-test-waiter.ts
@@ -21,7 +21,10 @@ export default class NoopTestWaiter implements ITestWaiter {
   waitUntil(): boolean {
     return true;
   }
+
   debugInfo(): ITestWaiterDebugInfo[] {
     return [];
   }
+
+  reset(): void {}
 }
--- a/addon/test-waiter.ts
+++ b/addon/test-waiter.ts
@@ -117,4 +117,14 @@ export default class TestWaiter<T = Token> implements ITestWaiter<T> {
   debugInfo(): ITestWaiterDebugInfo[] {
     return [...this.items.values()];
   }
+
+  /**
+   * Resets the waiter state, clearing items tracking async operations in this waiter.
+   *
+   * @public
+   * @method reset
+   */
+  reset(): void {
+    this.items.clear();
+  }
 }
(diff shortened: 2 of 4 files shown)


rwjblue:
I think we also need to add `.clear()` to the types:

https://github.com/rwjblue/ember-test-waiters/blob/873f7d577c34c47f33601b628e59564451a7c53d/addon/types/index.ts#L10

scalvert:
I thought about that, but I think we may want to keep the interface lean. We don't operate on an `ITestWaiter` anywhere, so unless there's a compelling reason to add it I think we should err on the side of keeping the interface simple. Otherwise, we should widen the interface with all the members, no?

scalvert:
I wonder if we should call this `reset` rather than `clear`, as we're resetting the waiter state?

scalvert:
Renamed to `reset`!

scalvert:
> I think we also need to add `.clear()` to the types:
> 
> https://github.com/rwjblue/ember-test-waiters/blob/873f7d577c34c47f33601b628e59564451a7c53d/addon/types/index.ts#L10

After a second look, it makes sense to add this.

.travis.yml
```diff
@@ -3,7 +3,7 @@ language: node_js
 node_js:
   # we recommend testing addons with the same minimum supported node version as Ember CLI
   # so that your addon works for all apps
-  - '6'
+  - '8'
```
bendemboski: It just occurred to me that we can solve this using yarn resolutions and stick with node 6 for now. If we add

```
"**/sort-package-json": "1.23.0"
```

the the resolutions section of package.json, the dependencies will all install on node 6.

bendemboski: Eh, never mind. This would technically work, but I realized that in my PR I'm going to need to build on node 8 anyway because to build against Ember 3.12+ I need to use the 1.0.0 of `ember-concurrency-decorators` and it doesn't support Node 6. So may as well jump to 8 here.

scalvert: Ya I think we can leave this for now. Resolutions is a good idea until you forget to remove them, which renders dependabot effectively useless for that dependency.

scalvert:
We need to make sure #86 is merged before this, so we can ensure this API exists for consumers.

PR #88, opened 2019-12-13
Converting to use github actions

scalvert:


config/ember-try.js
```diff
@@ -33,6 +33,14 @@ module.exports = function() {
             },
           },
         },
+        {
+          name: 'ember-lts-3.8',
```
rwjblue: Can you also add 3.12?

.travis.yml
```diff
@@ -1,72 +0,0 @@
----
```
rwjblue: Lets land the new thing and delete this file once we have confirmed it is working (after landing to master and having a few passing CI runs).

.github/workflows/nodejs.yml
```diff
@@ -0,0 +1,64 @@
+name: CI Build
+
+on: [push]
```
rwjblue: Need to add more things here, at least `pull_request` (or it won't run on PR's from a fork)

.github/workflows/nodejs.yml
```diff
@@ -0,0 +1,64 @@
...
+      - if: matrix.os == 'macOS-latest'
+        run: |
+          brew update
+          brew cask install google-chrome
+      - uses: actions/checkout@v1
+      - uses: actions/setup-node@v1
+        with:
+          node-version: 8.x
+      - name: install yarn
+        run: npm install -g yarn
```
Turbo87: yarn is being preinstalled from the `setup-node` action. this step is not needed most of the time

config/ember-try.js
scalvert: Done

.github/workflows/nodejs.yml
scalvert: Added `pull_request`.

.github/workflows/nodejs.yml
scalvert: Removed explicit installation of yarn. Thanks!

.travis.yml
scalvert: Added it back for now. Seems reasonable.

.travis.yml
```diff
@@ -3,7 +3,7 @@ language: node_js
 node_js:
   # we recommend testing addons with the same minimum supported node version as Ember CLI
   # so that your addon works for all apps
-  - '6'
+  - '8'
```
Turbo87: why did you change the Node.js version for TravisCI here?

scalvert: Because the old version of node is causing issues when testing floating dependencies. We're pulling in packages that don't support 6 in their engines anymore, so we have to do this anyway.

We've updated a few other PRs (yet unmerged) in this repo to use Node 8 in the .travis.yml. This PR will just preempt them.

Turbo87: then we should update https://github.com/rwjblue/ember-test-waiters/blob/master/package.json#L73 and declare it officially as a breaking change

scalvert: We don't want to do that yet until Node 8 is EOL, which is 12/31/19 (soon). Once that happens we will bump the engines version and major bump. We just wanted to avoid two major version bumps in succession. 

Hope this makes sense.

Turbo87: yeah! but then we should keep testing on Node 6

scalvert: So as I said, if we keep Node 6 as the target for travis, when the floating dependencies test runs, the `yarn install` fails due to some packages requiring Node 8 and above. This would have prevented us from merging this PR, as CI would have failed. Perhaps there's a better way to address this. What did you have in mind?

Turbo87: we could use Yarn resolutions to avoid this issue. not perfect, but better than not running CI on Node 6 at all.

scalvert: This seems like a reasonable alternative, but I want to emphasize that we're not planning on releasing until Node 8 is EOL. The current release is still tested on Node 6, and will continue to work for those users using Node 6.

I can totally use resolutions instead for this, but I'm not sure it's worth it at this point. If you feel strongly about it, I'm happy to open a new PR that updates both the travis and github actions implementations to use Node 6, and force the offending package to a version that supports Node 6 via resolutions. Let me know.

Turbo87: if that is the case then we should update the package.json file to reflect those new minimum requirements on master

scalvert: Yep that's fine. I can do that.

PR #95, opened 2020-01-10
Dropping support for node 8

scalvert:
This PR also removes the redundant .travis.yml.

### DIFF
--- a/.github/workflows/ci-build.yml
+++ b/.github/workflows/ci-build.yml
@@ -26,7 +26,7 @@ jobs:
       - uses: actions/checkout@v1
       - uses: actions/setup-node@v1
         with:
-          node-version: 8.x
+          node-version: 10.x
       - name: install dependencies
         run: yarn install
       - name: lint
--- a/package.json
+++ b/package.json
@@ -70,11 +70,8 @@
     "release-it-lerna-changelog": "^1.0.3",
     "typescript": "^3.7.4"
   },
-  "resolutions": {
-    "**/engine.io": "~3.3.0"
-  },
   "engines": {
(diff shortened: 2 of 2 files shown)


scalvert:
We should also wait for #96 to merge before releasing, so we can have nicer changelogs :)

scalvert:
Going to wait for #86 to be merged and released before merging and releasing this in a major version bump.

scalvert:
@rwjblue I think since #86 is merged we can think about merging this and releasing. Thoughts?

package.json
```diff
@@ -74,7 +74,7 @@
     "**/engine.io": "~3.3.0"
   },
   "engines": {
-    "node": "6.* || 8.* || >= 10.*"
+    "node": ">= 10.*"
```
rwjblue:
```suggestion
    "node": "10.* || >= 12.*"
```

package.json
```diff
@@ -74,7 +74,7 @@
     "**/engine.io": "~3.3.0"
```
rwjblue: I think we can remove this

.github/workflows/ci-build.yml
```diff
@@ -61,7 +61,7 @@ jobs:
       - uses: actions/checkout@v1
       - uses: actions/setup-node@v1
         with:
-          node-version: 12.x
```
rwjblue: Let's keep this using 12.x

package.json
scalvert: Done

.github/workflows/ci-build.yml
scalvert: Updated

Issue #110, opened 2020-02-06
Nit request: Make tokens always truthy


sclaxton:
https://github.com/rwjblue/ember-test-waiters/blob/1a09a527d6261385d390d056a2b985074b74fc97/addon/test-waiter.ts#L4

Can we have token enumeration start from `1`? Just got trolled checking if a token existed or not using a simple `if (token) {}` check and it didn't work cause it was `0`...

scalvert:
This seems entirely reasonable. It's hard to know that the properties of the first token are effectively different from the rest without knowing it's a number.

scalvert:
Closed via #111

PR #111, opened 2020-02-06
Fixes default tokens to be truthy

scalvert:
Currently in the `test-waiter` module, we use a zero-based counter for the default tokens. This presents a problem for consumers when checking for the existence of a valid token:

```js

// ...

if (token) {
  waiter.endAsync(token);
}
```

This PR changes tokens to be one-based, so that the truthy evaluation will always pass, and consumers won't have to know that the first token is magically different than the rest.


scalvert:
Fixes #111

scalvert:
Fixes #110

PR #112, opened 2020-02-07
Do not throw errors for endAsync being called multiple times

### LINKED ISSUE #4: Determine whether we need to track when `endAsync` is called more than once.

scalvert:
Fixes #4

### DIFF
--- a/addon/test-waiter.ts
+++ b/addon/test-waiter.ts
@@ -60,7 +60,7 @@ export default class TestWaiter<T extends object = Token> implements ITestWaiter
    * @param item {T} The item to register for waiting
    * @param label {string} An optional label to identify the item
    */
-  beginAsync(token: T = this.nextToken(), label?: string) {
+  beginAsync(token: T = this.nextToken(), label?: string): T {
     this.register();
 
     if (this.items.has(token)) {
@@ -81,16 +81,16 @@ export default class TestWaiter<T extends object = Token> implements ITestWaiter
 
   /**
    * Should be used to signal the end of an async operation. Invocation of this
-   * method should be paired with a preceding `beginAsync` call, which would indicate the
-   * beginning of an async operation.
+   * method should be paired with a preceding `beginAsync` call from this instance,
+   * which would indicate the beginning of an async operation.
    *
    * @public
    * @method endAsync
    * @param item {T} The item to that was registered for waiting
    */
-  endAsync(token: T) {
+  endAsync(token: T): void {
     if (!this.items.has(token) && !this.completedOperations.has(token)) {
-      throw new Error(`endAsync called for ${token} but it is not currently pending.`);
+      throw new Error(`endAsync called with no preceding beginAsync call.`);
     }
 
     this.items.delete(token);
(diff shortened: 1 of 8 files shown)


addon/test-waiter.ts
```diff
@@ -14,12 +13,14 @@ function getNextToken(): number {
...
  */
-export default class TestWaiter<T = Token> implements ITestWaiter<T> {
+export default class TestWaiter<T extends object | number = Token> implements ITestWaiter<T> {
   public name: WaiterName;
   private nextToken: () => T;
   private isRegistered = false;
 
   items = new Map<T, ITestWaiterDebugInfo>();
+  completedOperationsForTokens = new WeakMap<Token, boolean>();
+  completedOperationsForNumbers = new Map<number, boolean>();
```
rwjblue: Technically, this is for anything that is not an object or function right?

addon/test-waiter.ts
```diff
@@ -128,4 +132,10 @@ export default class TestWaiter<T = Token> implements ITestWaiter<T> {
   reset(): void {
     this.items.clear();
   }
+
+  private getCompletedOperations(token: T) {
```
rwjblue: Let's underscore this method (the TS `private` only helps us with TS consumers).

addon/test-waiter.ts
```diff
@@ -128,4 +132,10 @@ export default class TestWaiter<T = Token> implements ITestWaiter<T> {
...
+    return typeof token === 'number'
```
rwjblue: I think this should actually be:

```suggestion
    let type = typeof token;
    
    return token === null || (type !== 'function' && type !== 'object');
```

addon/test-waiter.ts
scalvert: It _is_, but I believe you and I discussed that we want to phase out allowing users to supply their own tokens, right? Based on that, this is intended to be a temporary solution to still maintain backwards compatibility with `number` types for tokens, while driving people over to using the `Token` type, which is specifically what we want them to use. Is this correct?

addon/test-waiter.ts
scalvert: Sure. Can do.

addon/test-waiter.ts
scalvert: I don't think so, as that would break the functionality (this is expecting to return a map-like object).

It looks like you're proposing to just invert the condition, to make it anything that's _not_ any of the above conditions, correct?

scalvert: I've updated it to look for everything but `function` and `object`.

addon/test-waiter.ts
scalvert: Discussed offline. We want to allow for primitives as token types, and as such I've updated the type to be `Primative`.

PR #117, opened 2020-02-14
Rename interfaces, remove class exports.

scalvert:
This PR addresses comments as part of [RFC 581](https://github.com/emberjs/rfcs/pull/581). Specifically, [it was requested to keep the naming standards for interfaces consistent with other core repositories](https://github.com/emberjs/rfcs/pull/581#issuecomment-586461582). As such, all preceding `I`s are removed from the interface names. 

To accommodate that change, the core `TestWaiter` class was renamed. **The new name is not final and is open to suggestion**.


API.md
```diff
@@ -9,18 +9,20 @@
   - [getPendingWaiterState][5]
   - [hasPendingWaiters][6]
   - [\_reset][7]
-- [Test Waiter class][8]
-  - [TestWaiter][9]
+- [Async Test Waiter class][8]
```
rwjblue: FWIW, I think we want to avoid mentioning the "class" bit here

addon/async-test-waiter.ts
```diff
@@ -13,12 +13,13 @@ function getNextToken(): Token {
  * @public
  * @class TestWaiter<T>
  */
-export default class TestWaiter<T extends object | Primitive = Token> implements ITestWaiter<T> {
+export default class AsyncTestWaiter<T extends object | Primitive = Token>
```
rwjblue: Let's call this `TestWaiterImpl` (IMHO adding "async" doesn't _really_ add any value), but rename the imported interface locally:

addon/async-test-waiter.ts
```diff
@@ -1,4 +1,4 @@
-import { ITestWaiter, ITestWaiterDebugInfo, Primitive, WaiterName } from './types';
+import { Primitive, TestWaiter, TestWaiterDebugInfo, WaiterName } from './types';
```
rwjblue: Thoughts on moving this implementation directly into `addon/build-waiter.ts`, and restructuring the class to be a non-default (possibly underscored) export?

addon/build-waiter.ts
```diff
@@ -1,5 +1,6 @@
+import { AsyncTestWaiter, TestWaiter } from 'ember-test-waiters';
```
rwjblue: I'd love it if we just move the class directly into here, simplifies the "mental model" and helps us avoid "exposing" the class when we don't really want to.

addon/index.ts
```diff
@@ -8,7 +8,7 @@ export {
 } from './waiter-manager';
 
 export { default as Token } from './token';
-export { default as TestWaiter } from './test-waiter';
+export { default as AsyncTestWaiter } from './async-test-waiter';
```
rwjblue: I'd like to **NOT** export this class _at all_...

addon/noop-test-waiter.ts
```diff
@@ -1,4 +1,4 @@
-import { ITestWaiter, ITestWaiterDebugInfo } from './types';
+import { TestWaiter, TestWaiterDebugInfo } from './types';
```
rwjblue: Same here, lets move this into `build-waiter.ts` and have that be the _only_ external importable thing

addon/wait-for-promise.ts
```diff
@@ -1,7 +1,7 @@
+import AsyncTestWaiter from './async-test-waiter';
 import { DEBUG } from '@glimmer/env';
-import TestWaiter from './test-waiter';
 
-const PROMISE_WAITER = new TestWaiter<Promise<unknown>>('promise-waiter');
+const PROMISE_WAITER = new AsyncTestWaiter<Promise<unknown>>('promise-waiter');
```
rwjblue: We should be using our own public API here, let's swap this to `buildWaiter()`

documentation.yml
```diff
@@ -7,9 +7,9 @@ toc:
       - getPendingWaiterState
       - hasPendingWaiters
       - _reset
-  - name: Test Waiter class
+  - name: Async Test Waiter class
```
rwjblue: I don't want to reference the "class" at all here, I think just calling it `Test Waiter` would be perfect...

tests/unit/async-test-waiter-test.ts
```diff
@@ -1,4 +1,10 @@
-import { getPendingWaiterState, getWaiters, TestWaiter, Token, _reset } from 'ember-test-waiters';
+import {
```
rwjblue: We should move these tests to directly use `buildWaiter`

API.md
scalvert: 👍

addon/async-test-waiter.ts
scalvert: Uggg. That feels so Java-y. It's offending my sensibilities. I'll do it thought since we're planning on essentially making it private-ish. I don't care as much about the name here, as it's not public.

addon/async-test-waiter.ts
scalvert: Ya I like this. I'll do it with the other changes.

addon/build-waiter.ts
scalvert: Agree.

addon/index.ts
scalvert: I'm with you there. This would be a breaking change though, no?

addon/noop-test-waiter.ts
scalvert: Well thees are the types. I think we do want to support these types being exported, no?

addon/wait-for-promise.ts
scalvert: Good call.

documentation.yml
scalvert: Yep, seems fine.

tests/unit/async-test-waiter-test.ts
scalvert: Can do.

addon/noop-test-waiter.ts
scalvert: Wait, I misunderstood. Thought you were referring to exporting the types from the root. I agree with this.

API.md
scalvert: This is actually removed from the docs now. Since we're explicitly documenting publicly exported stuff, this is now gone.

documentation.yml
scalvert: It's gone now, as we're explicitly documenting public exports now.

addon/build-waiter.ts
scalvert: Done.

addon/build-waiter.ts
```diff
@@ -1,14 +1,194 @@
...
+import { register } from './waiter-manager';
+
+function getNextToken(): Token {
+  return new Token();
+}
+
+/**
+ * A class providing creation, registration and async waiting functionality.
+ *
+ * @public
```
rwjblue: I think this should be private

```suggestion
 * @private
```

addon/build-waiter.ts
```diff
@@ -1,14 +1,194 @@
...
+ * @class TestWaiterImpl<T>
```
rwjblue: Do we also have to state the name again here?

addon/build-waiter.ts
```diff
@@ -1,14 +1,194 @@
...
+ */
+export class TestWaiterImpl<T extends object | Primitive = Token> implements TestWaiter<T> {
+  public name: WaiterName;
+  private nextToken: () => T;
+  private isRegistered = false;
+
+  items = new Map<T, TestWaiterDebugInfo>();
+  completedOperationsForTokens = new WeakMap<Token, boolean>();
+  completedOperationsForPrimitives = new Map<Primitive, boolean>();
```
rwjblue: Can you prefix all of these with `private` (in TS) and prefix with an _ to make it more clear for folks that stumble?

scalvert: For surely.

addon/build-waiter.ts
rwjblue: Resolving for now, we are going to have to revisit the documentation generation soon but not in this PR

addon/build-waiter.ts
rwjblue: After discussion, we are going to hold off on this for now.

addon/index.ts
```diff
@@ -7,9 +7,8 @@ export {
   hasPendingWaiters,
 } from './waiter-manager';
 
-export { default as Token } from './token';
-export { default as TestWaiter } from './test-waiter';
 export { default as buildWaiter } from './build-waiter';
 export { default as waitForPromise } from './wait-for-promise';
+export { default as Token } from './token';
```
rwjblue: We probably don't need to export this either, perhaps we could make a `buildToken()` helper if we really do want other folks to use the same kind of token...

addon/build-waiter.ts
scalvert: Negativo

tests/unit/build-waiter-test.ts
```diff
@@ -11,44 +12,44 @@ module('test-waiter', function(hooks) {
...
+    let waiter = buildWaiter('my-waiter');
 
     let token = waiter.beginAsync();
 
     let registeredWaiters = getWaiters();
 
     assert.equal(registeredWaiters[0], waiter, 'The waiter is registered');
     assert.deepEqual(
-      (<TestWaiter>registeredWaiters[0]).items.keys().next().value,
+      (<TestWaiterImpl>registeredWaiters[0]).items.keys().next().value,
```
rwjblue: Hmmm, I'd rather cast than export the class just for this purpose...

tests/unit/waiter-manager-noop-test.ts
```diff
@@ -1,7 +1,8 @@
+import { _reset, buildWaiter, getWaiters, register } from 'ember-test-waiters';
 import { module, test } from 'qunit';
+
 import { DEBUG } from '@glimmer/env';
-import { _reset, buildWaiter, register, getWaiters } from 'ember-test-waiters';
-import NoopTestWaiter from 'ember-test-waiters/noop-test-waiter';
+import NoopTestWaiter from 'ember-test-waiters/build-waiter';
```
rwjblue: This seems wrong? I don't think it is the default export of that file right? Maybe it isn't used at all?

addon/index.ts
scalvert: Fair. I think `Token` needs an exported type.

scalvert: Discussed offline: going to export `type Token = unknown`.

tests/unit/build-waiter-test.ts
scalvert: Cast to what? The interface doesn't have an items prop.

tests/unit/waiter-manager-noop-test.ts
scalvert: No it's not. Not sure how this isn't erroring out. Fixed now.

tests/unit/build-waiter-test.ts
rwjblue:
```ts
let waiter = registeredWaiters[0] as any;
assert.deepEqual(waiter.items.keys().next().value, whateverHere)
```

rwjblue: We should refactor these tests to leverage the debug info apis instead of accessing `.items` privately, once we do that the nastiness of the any cast can go away...

scalvert: In this case this test is explicitly looking for a token's presence. We don't expose the token via our debug info, so that won't work. I've resigned to cast to `any` for now until we can clean this up later.

PR #126, opened 2020-03-07
Adding constraints around naming test waiters

scalvert:
This PR adds some constrains around the naming of test waiters. Specifically, it adds a pattern (`namespace:descriptor`) for test waiters in order to help identify them. Currently this is implemented as a warning, to avoid a breaking change.

[README](https://github.com/emberjs/ember-test-waiters/blob/waiter-label-naming/README.md) changes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,6 +25,7 @@ This addon implements the design specified in [RFC 581](https://github.com/ember
 - [Installation](#installation)
 - [Quickstart](#quickstart)
   - [buildWaiter function](#buildwaiter-function)
+    - [Waiter naming conventions](#waiter-naming-conventions)
   - [waitForPromise function](#waitforpromise-function)
   - [Waiting on all waiters](#waiting-on-all-waiters)
 - [General Design](#general-design)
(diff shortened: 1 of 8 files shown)


addon/build-waiter.ts
```diff
@@ -3,6 +3,9 @@ import { Primitive, TestWaiter, TestWaiterDebugInfo, WaiterName } from 'ember-te
 import { DEBUG } from '@glimmer/env';
 import Token from './token';
 import { register } from './waiter-manager';
+import { warn } from '@ember/debug';
+
+const WAITER_NAME_PATTERN = /^[^:]*:.*/;
```
rwjblue: I think this needs to be tweaked a bit, to ensure the patterns listed in the README work nicely. Specifically, these two:

> 1. `file-name` - if your file has only one waiter
> 1. `addon-name` - if your addon has only one waiter

Neither of these have a `:`, and right now the regexp would cause them to warn.

addon/build-waiter.ts
```diff
@@ -134,8 +137,19 @@ class NoopTestWaiter implements TestWaiter {
  * }
  */
 export default function buildWaiter(name: string): TestWaiter {
```
rwjblue: Can you add a warning when using the same name twice?

```ts
const WAITER_NAMES = DEBUG ? new Set() : undefined;
export default function buildWaiter(name: string): TestWaiter {
  if (DEBUG) {
    warn('dont use the same name twice dummy', !WAITER_NAMES.has(name), { id: '@ember/test-waiters.duplicate-waiter-name' });
    WAITER_NAMES.add(name);
  }

  // ...snip...
}
```

addon/build-waiter.ts
scalvert: Yep. Good catch. Will add `?` for optional `:`.

addon/build-waiter.ts
scalvert: Yep. I specifically like the usage of 'dummy' in there.

addon/build-waiter.ts
scalvert: Updated.

addon/build-waiter.ts
scalvert: Added. Good suggestion.

scalvert:
@rwjblue bump.

README.md
```diff
@@ -25,6 +25,7 @@ This addon implements the design specified in [RFC 581](https://github.com/ember
 - [Installation](#installation)
 - [Quickstart](#quickstart)
   - [buildWaiter function](#buildwaiter-function)
+    - [Waiter naming conventions](#waiter-naming-conventions)
```
rwjblue, line 28: Should this be down in the best practices section, with a link from here?

README.md
```diff
@@ -59,17 +60,31 @@ that provides a number of methods. The key methods that allow you to control asy
 a pair to _begin_ waiting and _end_ waiting respectively. The `beginAsync` method returns a `token`, which uniquely identifies that async operation. To mark the
 async operation as complete, call `endAsync`, passing in the `token` that was returned from the prior `beginAsync` call.
 
+#### Waiter naming conventions
```
rwjblue, line 63: Can you add a link to the `eslint-plugin-ember` rule docs in this section too?

Issue #129, opened 2020-04-03
TypeScript: regular promises not compatible with RSVP?


NullVoxPopuli:
Tried upgrading to 2.0.0 again, but am getting these type errors:

```ts
app/services/toast.ts:28:20 - error TS2345: Argument of type 'Promise<unknown>' is not assignable to parameter of type 'import("C:/Users/Preston/Documents/Development/NullVoxPopuli/emberclear/packages/frontend/node_modules/@types/rsvp/index").Promise<unknown>'.
  Property 'new' is missing in type 'Promise<unknown>' but required in type 'import("C:/Users/Preston/Documents/Development/NullVoxPopuli/emberclear/packages/frontend/node_modules/@types/rsvp/index").Promise<unknown>'.

28     waitForPromise(createToast(status, message, options));
                      ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

  node_modules/@types/rsvp/index.d.ts:103:9
    103         new<T>(
                ~~~
    'new' is declared here.
```

whether or not I have `@types/rsvp` installed myself seems to not matter.

I'm not sure why I'm getting an RSVP error now? unless ember-test-waiters uses TS and is using an old version of `@types/rsvp`? 🤔

scalvert:
Thanks, @NullVoxPopuli. I could be wrong, but I think the issue here is you're trying to intermix native and RSVP promises.

https://github.com/NullVoxPopuli/emberclear/blob/master/packages/frontend/app/services/toast.ts#L57

You're not importing RSVP promise in this service, but rather using native promises, so `waitForPromise` will error due to a type mismatch.

NullVoxPopuli:
> waitForPromise will error due to a type mismatch.

right, but waitForPromise is compat with native Promises, so the types should be updated, yeah?

https://github.com/emberjs/ember-test-waiters/pull/130

scalvert:
We specifically [import `RSVP.Promise`](https://github.com/emberjs/ember-test-waiters/blob/master/addon/wait-for-promise.ts#L2) and use that in `waitForPromise`. So I'm not sure that's true?

scalvert:
We additionally specifically use `RSVP.Promise` [in our tests](https://github.com/emberjs/ember-test-waiters/blob/master/tests/unit/wait-for-promise-test.ts#L6), so we're not testing for native promises.

NullVoxPopuli:
[...]

PR #130, opened 2020-04-03
Fix Type Errors

NullVoxPopuli:
Related so #129

All 5 of the type errors introduced by ember-test-waiters 2.0.0 are resolved by this PR.
https://github.com/NullVoxPopuli/emberclear/pull/558
image

### DIFF
--- a/addon/wait-for-promise.ts
+++ b/addon/wait-for-promise.ts
@@ -4,6 +4,12 @@ import buildWaiter from './build-waiter';
 
 const PROMISE_WAITER = buildWaiter('promise-waiter');
 
+type PromiseType<T> = Promise<T> | RSVP.Promise<T>;
+
+interface Thenable<T, Return extends PromiseType<T>> {
+  then(resolve: (value: T) => T, reject?: (error: Error) => T): Return;
+}
+
 /**
  * A convenient utility function to simplify waiting for a promise.
  *
@@ -25,26 +31,25 @@ const PROMISE_WAITER = buildWaiter('promise-waiter');
  *   }
  * }
  */
-export default function waitForPromise<T, PromiseType extends Promise<T> | RSVP.Promise<T>>(
-  promise: PromiseType,
+export default function waitForPromise<T, KindOfPromise extends PromiseType<T>>(
+  promise: KindOfPromise,
   label?: string
-): PromiseType {
+): KindOfPromise {
   let result = promise;
 
   if (DEBUG) {
     PROMISE_WAITER.beginAsync(promise, label);
 
-    result = (promise as any) // sorry
-      .then(
-        (value: T) => {
-          PROMISE_WAITER.endAsync(promise);
-          return value;
-        },
-        (error: Error) => {
-          PROMISE_WAITER.endAsync(promise);
-          throw error;
-        }
-      );
+    result = ((promise as unknown) as Thenable<T, KindOfPromise>).then(
+      (value: T) => {
+        PROMISE_WAITER.endAsync(promise);
+        return value;
+      },
+      (error: Error) => {
+        PROMISE_WAITER.endAsync(promise);
+        throw error;
+      }
+    );
   }
 
   return result;
--- a/addon/wait-for-promise.ts
+++ b/addon/wait-for-promise.ts
@@ -1,5 +1,5 @@
 import { DEBUG } from '@glimmer/env';
-import { Promise as RSVPPromise } from 'rsvp';
+import RSVP from 'rsvp';
 import buildWaiter from './build-waiter';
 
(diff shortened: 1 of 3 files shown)


addon/wait-for-promise.ts
```diff
@@ -25,18 +25,21 @@ const PROMISE_WAITER = buildWaiter('promise-waiter');
...
+export default function waitForPromise<T>(
+  promise: Promise<T> | RSVPPromise<T>,
+  label?: string
+): Promise<T> | RSVPPromise<T> {
   let result = promise;
 
   if (DEBUG) {
     PROMISE_WAITER.beginAsync(promise, label);
 
     result = promise.then(
```
NullVoxPopuli, line 37: currently have 
image

addon/wait-for-promise.ts
```diff
@@ -25,22 +25,26 @@ const PROMISE_WAITER = buildWaiter('promise-waiter');
  *   }
  * }
  */
-export default function waitForPromise<T>(promise: Promise<T>, label?: string): Promise<T> {
+export default function waitForPromise<T, PromiseType extends Promise<T> | RSVPPromise<T>>(
```
scalvert, line 28: This looks good. I looked into us potentially using `PromiseLike<T>`, but using that would narrow the return type (it wouldn't include `catch` and `finally`). 

Could we add tests for both variants, native and RSVP?

NullVoxPopuli: yea!

NullVoxPopuli:
@scalvert tests added!

addon/waiter-manager.ts
```diff
@@ -53,6 +53,10 @@ export function getWaiters(): Waiter[] {
  * @private
  */
 export function _reset(): void {
+  for (let waiter of getWaiters()) {
```
NullVoxPopuli, line 56: this fixes a bug where if there are multiple calls to `waitForPromise`, `getPendingWaiterState()` _always_ returns no waiters. Because, `PROMISE_WAITER` is defined in module-space, is is not garbage collected during `WAITERS.clear()` (below). Re-setting `isRegistered` to false here allows the `PROMISE_WAITER` to be re-registered in the `WAITERS` `Map` during the next test.

scalvert: Nice find. Thanks for this.

addon/wait-for-promise.ts
```diff
@@ -1,14 +1,20 @@
 import { DEBUG } from '@glimmer/env';
-import { Promise } from 'rsvp';
+import RSVP from 'rsvp';
 import buildWaiter from './build-waiter';
 
 const PROMISE_WAITER = buildWaiter('promise-waiter');
 
+type PromiseType<T> = Promise<T> | RSVP.Promise<T>;
+
+interface Thenable<T, Return extends PromiseType<T>> {
```
scalvert, line 9: Why did we need to add this? Also, can we not use `PromiseLike<T>` in this case, as it's basically a thenable interface?

NullVoxPopuli: `PromiseLike<T>` doesn't have `catch`
image

`Thenable` doesn't either, but it doesn't need to, because `PromiseType` does have `catch`. :shrug:

`PromiseLike<T>` also for some reason isn't compatible with `PromiseType` like `Thenable` is. :-\

idk. I don't like it, but I don't know a better way, and TS isn't mature enough yet to lead me to a better way :(

scalvert: Ya I sorta knew this was the reason, as your `Thenable` implementation allows you to specify the `T` (value) _and_ `Return` (promise type, whether `Promise` or `RSVP.Promise`), whereas `PromiseLike` doesn't allow you to do that.

It's fuggly (as you said), but does work. It's a bit irritating to have to jump through these sorts of hooks, but that's the cost of typing I guess.

addon/wait-for-promise.ts
scalvert: Thank you!

addon/wait-for-promise.ts
NullVoxPopuli: Def the cost of typing when we're only sprinkling types on top of untyped stuff. 
if stuff were typed to begin with, everything would be much smoother :D

rwjblue: > if stuff were typed to begin with, everything would be much smoother

agree to disagree :smile:

NullVoxPopuli: I don't think that's a disagreement? :) Idk
It's more 'in addition to' I guess.

Like, in natively typed languages i've had a much easier time resolving type errors.

PR #131, opened 2020-04-16
Extend waitForPromise to work in function form

bendemboski:
Extracted from #81 

A couple of notes:
* I have not address https://github.com/emberjs/ember-test-waiters/pull/81#discussion_r353930902 because I'm still a typescript noob and a jsdoc noob (so noob squared here)
* I had to do some interesting acrobatics to keep the `waitForPromise` tests reasonably DRY -- I'm definitely open to other ideas if anyone has any

### DIFF
--- a/addon/ember-test-waiters/types/index.ts
+++ b/addon/ember-test-waiters/types/index.ts
@@ -1,3 +1,5 @@
+import RSVP from 'rsvp';
+
 /**
  * @type WaiterName
  *
@@ -105,4 +107,10 @@ export interface PendingWaiterState {
   waiters: {
     [waiterName: string]: TestWaiterDebugInfo[] | true;
   };
+}
+
+export type PromiseType<T> = Promise<T> | RSVP.Promise<T>;
+
+export interface Thenable<T, Return extends PromiseType<T>> {
+  then(resolve: (value: T) => T, reject?: (error: Error) => T): Return;
 }
--- a/addon/ember-test-waiters/wait-for.ts
+++ b/addon/ember-test-waiters/wait-for.ts
@@ -1,0 +1,100 @@
+import { DEBUG } from '@glimmer/env';
+import waitForPromise from './wait-for-promise';
+
+type AsyncFunction<A extends Array<any>, PromiseReturn> = (...args: A) => Promise<PromiseReturn>;
+type AsyncFunctionArguments = [AsyncFunction<any[], any>, string?];
+type DecoratorArguments = [object, string, PropertyDescriptor, string?];
+
+/**
+ * A convenient utility function to simplify waiting for a promise. Can be used
+ * in both decorator and function form, and the function form either accepts a
+ * promise or an async function.
+ *
+ * @public
+ * @param promise {Promise<T> | RSVP.Promise<T>} The promise to track async operations for
+ * @param label {string} An optional string to identify the promise
+ *
+ * @example
+ *
+ * import Component from '@ember/component';
+ * import { waitForPromise } from '@ember/test-waiters';
+ *
+ * export default class Friendz extends Component {
+ *   didInsertElement() {
+ *     waitForPromise(new Promise(resolve => {
+ *       doSomeWork();
+ *       resolve();
+ *     }));
+ *   }
+ * }
+ *
+ * @public
+ * @param promise {Function} An async function
+ * @param label {string} An optional string to identify the promise
+ *
+ * @example
+ *
+ * import Component from '@ember/component';
+ * import { waitForPromise } from 'ember-test-waiters';
+ *
+ * export default Component.extend({
+ *   doAsyncStuff: waitForPromise(async function doAsyncStuff() {
+ *     await somethingAsync();
+ *   }
+ * });
+ *
+ * @example
+ *
+ * import Component from '@ember/component';
+ * import { waitForPromise } from 'ember-test-waiters';
+ *
+ * export default class Friendz extends Component {
+ *   @waitForPromise
+ *   async doAsyncStuff() {
+ *     await somethingAsync();
+ *   }
+ * }
+ *
+ */
+export default function waitFor(fn: AsyncFunction<any[], any>, label?: string): Function;
+export default function waitFor(
+  target: object,
+  _key: string,
+  descriptor: PropertyDescriptor,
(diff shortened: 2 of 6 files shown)


scalvert:
@bendemboski - @rwjblue and I revived this one. Thanks so much for taking it as far as you did!

We've made a few changes, namely kept the `waitForPromise` API as it was, and added the `waitFor` API to provide the async function/decorator support. I think this will be a really nice addition to the addon.

Let me know if you have any thoughts/suggestions.

bendemboski:
LGTM. I think the two additional items in the original PR that are probably mostly ready (I'll have to revive them from my repo's branches) are:

1. babel transform to strip `waitForPromise()` and `waitFor()` from the build
2. `waitForCoroutine()`

Any changes on those fronts, or continue with the original plans?

scalvert:
I think the idea is that we layer on generator support into `waitFor`. Let me sync with @rwjblue after this is merged and we'll articulate the changes in an issue. I'll make sure to @ you on it.

As for the stripping, that's specifically to strip decorators?

bendemboski:
> I think the idea is that we layer on generator support into waitFor. Let me sync with @rwjblue after this is merged and we'll articulate the changes in an issue. I'll make sure to @ you on it.

Oh, gotcha, that makes sense from an API standpoint. And I think I can maybe picture how to implement it? We'd need to make the [wrapper function](https://github.com/emberjs/ember-test-waiters/pull/131/files#diff-866a715cbcb7a5ea49100304bf275319R74) look at the return value, and if it looks like an iterator, then wrap it in our own iterator instrumented with test waiter calls, instead of wrapping it in `waitForPromise()`, and return that...I think that would work?

> As for the stripping, is that so we can reduce the size of the bundle?

Yeah, this was originally something requested by @rwjblue and my cut at it was [here](https://github.com/emberjs/ember-test-waiters/pull/81/files#diff-5c5749580a4927862aeb6adfc15996fb).

scalvert:
@bendemboski yes that's exactly what we were thinking WRT the generator - essentially inspecting the shape of the return value to determine if it's an iterator, and wrapping it as you said.

As for stripping, I think we can wait on that. I think the idea we had was to still ship these things in the prod bundle, as we ultimately want to be able to run tests against a prod bundle, but still have it wait for async. If we strip, we can't do that. 

This is predicated on us moving away from using the `DEBUG` flag, but something intermediate to evaluate whether we're in test mode or not.

rwjblue:
> As for stripping, I think we can wait on that. I think the idea we had was to still ship these things in the prod bundle, as we ultimately want to be able to run tests against a prod bundle, but still have it wait for async. 

Ya, this just means we need another flag (`DEBUG` is not good enough here, because we need `ember test --environment=production` to be able to work.

bendemboski:
Gotcha. Okay, once this is finalized/merged I'll make some time to port the `waitForCoroutine()` implementation into `waitFor()`.

addon/ember-test-waiters/wait-for.ts
```diff
@@ -0,0 +1,81 @@
...
+    let [fn, label] = args as AsyncFunctionArguments;
+
+    if (!DEBUG) {
+      return fn;
+    }
+
+    return function(this: any, ...args: any[]) {
+      let promise = fn.call(this, ...args);
+
+      return waitForPromise(promise, label);
```
rwjblue, line 63:
```js
function isThenable(maybePromise: any | PromiseLike<unknown>): maybePromise is PromiseLike<unknown> {
    let type = typeof maybePromise;

    return ((maybePromise !== null && type === 'object') || type === 'function') && typeof maybePromise.then === 'function';
}
```

```js
      let result = fn.call(this, ...args);

      if (isThenable(result)) {
        return waitForPromise(result, label);
      } else {
        return result;
      }
```

addon/ember-test-waiters/wait-for.ts
```diff
@@ -0,0 +1,81 @@
...
+
+    if (!DEBUG) {
+      return descriptor;
+    }
+
+    let fn = descriptor.value;
+
+    descriptor.value = function(...args: any[]) {
+      let promise = fn.call(this, ...args);
+      return waitForPromise(promise, label);
```
rwjblue, line 76:
```suggestion
      let result = fn.call(this, ...args);

      if (isThenable(result)) {
        return waitForPromise(result, label);
      } else {
        return result;
      }
```

addon/ember-test-waiters/types/index.ts
```diff
@@ -106,3 +108,9 @@ export interface PendingWaiterState {
     [waiterName: string]: TestWaiterDebugInfo[] | true;
   };
 }
+
+export type PromiseType<T> = Promise<T> | RSVP.Promise<T>;
```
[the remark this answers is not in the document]
rwjblue: Chatted with @scalvert separately about this, we'll land this as is (since its just moving this code around) and will fix it in a stand alone PR...

addon/ember-test-waiters/wait-for.ts
```diff
@@ -0,0 +1,81 @@
...
+export default function waitFor(
+  target: object,
+  _key: string,
+  descriptor: PropertyDescriptor,
+  label?: string
+): PropertyDescriptor;
+export default function waitFor(
+  ...args: AsyncFunctionArguments | DecoratorArguments
+): PropertyDescriptor | Function {
+  let isAsyncFunction = args.length < 3;
```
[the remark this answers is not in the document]
scalvert: We've agreed to skip this for now.

addon/ember-test-waiters/wait-for.ts
```diff
@@ -71,11 +64,35 @@ export default function waitFor(
...
+  }
+
+  return function(this: any, ...args: any[]) {
+    let result = fn.call(this, ...args);
+
+    if (isThenable(result)) {
+      return waitForPromise(result, label);
+    } else {
+      return result;
+    }
```
rwjblue, line 85: Note: @bendemboski I think we would add the check for iterator here just like the `isThenable` thing (iterator being something with a `typeof result.next === 'function'` I believe)

scalvert: Ya maybe add an `isIterable`?

PR #132, opened 2020-04-16
Strip waitForPromise from production builds

bendemboski:
Extracted from #81

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -23,6 +23,7 @@ module.exports = {
         '.eslintrc.js',
         '.prettierrc.js',
         '.template-lintrc.js',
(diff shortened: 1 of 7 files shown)


scalvert:
There's an eventual plan for us to be able to run tests in prod (on a prod build). If we do this, we lose the ability to do this in a deterministic way. 

This will of course require a change to the use of `DEBUG` - we'd need a different mode we could key off.

I think we should discuss this more. 

cc @rwjblue

scalvert:
@bendemboski and I spoke offline about this. In light of us wanting to add support for running tests in prod builds, I'm going to close this. We can revisit  at a later time if that path changes.

PR #133, opened 2020-04-20
Expose `@ember/test-waiters` in addition to `ember-test-waiters` module names

scalvert:
Module rename as part of [RFC 0581](https://github.com/emberjs/rfcs/blob/master/text/0581-new-test-waiters.md)


addon/ember-test-waiters/index.js
```diff
@@ -0,0 +1 @@
+export * from '@ember/test-waiters';
```
rwjblue, line 1: We should add an issue to add a deprecation here I think.

addon/@ember/test-waiters/build-waiter.ts
```diff
@@ -1,4 +1,4 @@
-import { Primitive, TestWaiter, TestWaiterDebugInfo, WaiterName } from 'ember-test-waiters';
+import { Primitive, TestWaiter, TestWaiterDebugInfo, WaiterName } from '@ember/test-waiters';
```
rwjblue, line 1: Can you update this to be a relative path?

index.js
```diff
@@ -10,4 +10,8 @@ module.exports = {
 
     forceHighlander(this.project);
   },
+
+  treeForAddon(tree) {
+    return this.processedAddonJsFiles(tree);
+  },
```
rwjblue, line 16: Can you add a comment RE: not calling super (and not supporting templates or styles) and explaining what we are actually doing here?

addon/ember-test-waiters/index.js
scalvert: Agree.

index.js
scalvert: 👍

addon/@ember/test-waiters/build-waiter.ts
scalvert: Yep.

scalvert:
@rwjblue added a sanity check task that imports from `ember-test-waiters` to ensure the legacy module import works.

rwjblue:
@scalvert - Changes seem good, should we also change the package name while we are at it?

scalvert:
I removed the deprecation and updated some other references to the old module name. I also extended the highlander check to include the new module name.

Can you take another look, @rwjblue?

README.md
```diff
@@ -45,12 +45,12 @@ This addon implements the design specified in [RFC 581](https://github.com/ember
 ## Installation
 
 ```
-ember install ember-test-waiters
+ember install @ember/test-waiters
```
rwjblue, line 48: I don't think this part will work just yet, probably should revert back:

```suggestion
ember install ember-test-waiters
```

package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "ember-test-waiters",
+  "name": "@ember/test-waiters",
```
rwjblue, line 2:
```suggestion
  "name": "ember-test-waiters",
```

PR #135, opened 2020-05-12
Switch to using `ember-cli-version-checker` for gathering addons

scalvert:
Removes the use of a custom recursive search for discovering all addons in favor of using `ember-cli-version-checker`.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -42,7 +42,7 @@ module.exports = {
       },
       plugins: ['node'],
       rules: Object.assign({}, require('eslint-plugin-node').configs.recommended.rules, {
-        // add your custom rules and overrides for node files here
+        'node/no-unpublished-require': 'off',
       }),
     },
     {
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -41,9 +41,7 @@ module.exports = {
         node: true,
       },
       plugins: ['node'],
-      rules: Object.assign({}, require('eslint-plugin-node').configs.recommended.rules, {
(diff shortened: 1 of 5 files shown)


.eslintrc.js
```diff
@@ -42,7 +42,7 @@ module.exports = {
       },
       plugins: ['node'],
       rules: Object.assign({}, require('eslint-plugin-node').configs.recommended.rules, {
-        // add your custom rules and overrides for node files here
+        'node/no-unpublished-require': 'off',
```
rwjblue, line 45: Why?

scalvert: This is a bit heavy handed, but this triggered the no-unpublished-require rule because ember-cli-version-checker is a devDependency, but is published. I should likely either target the override to just this file, or move the dependency to dependencies in package.json.

package.json
```diff
@@ -48,6 +48,7 @@
     "ember-cli-test-loader": "^3.0.0",
     "ember-cli-typescript-blueprints": "^3.0.0",
     "ember-cli-uglify": "^3.0.0",
+    "ember-cli-version-checker": "^5.1.0",
```
rwjblue, line 51: This should be in `dependencies` not `devDependencies` (this is likely also why the eslint rule is failing)

.eslintrc.js
scalvert: Fixed.

package.json
scalvert: Fixed.

Issue #142, opened 2020-05-16
Missing type definitions for `@ember/test-waiters` 


NullVoxPopuli:
I just upgraded and switched my waitForPromise import from `@ember/test-helpers` and now I have a bunch of type errors.
![image](https://user-images.githubusercontent.com/199018/82121345-43e2a880-975a-11ea-9f1c-0b9628305043.png)

Is there a type package somewhere? or are the plans for this package to include its own types?

scalvert:
Thanks for bringing this to our attention, @NullVoxPopuli. 

This package _should_ be correctly publishing its own types, by re-exporting the types from `@ember/test-waiters` via the `ember-test-waiters` package. I'll need to look into it to see why they're not correctly re-exporting.

rwjblue:
Ya, most likely because we are still publishing under ember-test-waiters. For the new module paths to properly discover types we’d need to be using `@ember/test-helpers`. 

@NullVoxPopuli - In the meantime, I think you can work around using `paths` in your `tsconfig.json`

scalvert:
We identified the issue, I've implemented a fix and will push a PR up soon. Sorry for the issue, @NullVoxPopuli! And thanks again for bringing this to our attention.

rwjblue:
#148 fixed the issue folks were having when still using the `ember-test-waiters` import paths, it does not (and cannot) fix types for folks using `@ember/test-waiters`. In order to fix that we need to actually ship a release of the new package name (essentially reverting #148).

scalvert:
I believe this was resolved by #150

PR #148, opened 2020-05-18
Inverting module export locations for multi-module publishing

scalvert:
Fix for #142


rwjblue:
Note for others, this fixes the type issues that folks still using `ember-test-waiters` as their import path were hitting. This does **not** fix any TS issues when attempting to use `@ember/test-waiters` (there really isn't anything we can do about that issue until we release the actual `@ember/test-waiters` package).

PR #150, opened 2020-05-29
Converting published package name to @ember/test-waiters

scalvert:
Continuation of #133, #141 and #148. Final preparation for publishing under a new module name.

- Moves files back to `@ember/test-waiters` directory from `ember-test-waiters`
- Renames module references to `@ember/test-waiters`
- Adds deprecation for users who import from `ember-test-waiters` module


package.json
```diff
@@ -12,8 +12,8 @@
     "doc": "doc",
     "test": "tests"
   },
-  "main": "addon/@ember/test-helpers/index.ts",
-  "types": "addon/@ember/test-helpers/index.d.ts",
```
scalvert, line 16: Noticed this bug, which may have contributed to users not being able to correctly reference types in TypeScript.

simonihmig:
Due to unfortunate timing, I just tried to install this, using the new scoped module name according to the Readme, but that failed due to the new package not being published yet apparently. Is it expected to be released soon-ish? Or should I still use the old package for now?

scalvert:
Hey @simonihmig,

Sorry for the confusion. We're working on a few incremental changes to this package before it's published under the new name. 

You can use the old module name with new import paths, as we build both modules from the old package name as a bridge to this new package name. That will make moving to the new package a change only to your package.json. 

Hope this helps.

rwjblue:
We'll hopefully get the new package published in the next day or so (at least a beta). I'm a bit waylaid due to lack of power/internet, hopefully should be back in action by EOD tomorrow...

addon/ember-test-waiters/index.ts
```diff
@@ -1,22 +1,12 @@
...
-  register,
-  unregister,
-  getWaiters,
-  _reset,
-  getPendingWaiterState,
-  hasPendingWaiters,
-} from './waiter-manager';
+deprecate(
+  'Importing from ember-test-waiters is deprecated. Please import from @ember/test-waiters',
+  true,
```
Turbo87, line 5: https://api.emberjs.com/ember/3.26/functions/@ember/debug/deprecate says:

> `test` **Boolean** A boolean. If falsy, the deprecation will be displayed.

are we sure that this is working as intended? 😅

scalvert: Uggg, I really dislike the deprecate APIs...

PR #173, opened 2020-08-08
feat(wait-for) support coroutines/generators

bendemboski:
Make waitFor() also work on coroutine/generator functions such as those used with ember-concurrency

replaces #81


addon/@ember/test-waiters/wait-for.ts
```diff
@@ -96,3 +135,33 @@ function isThenable(
...
+  return {
+    next() {
+      try {
+        let val = generator.next();
+        if (val.done) {
+          GENERATOR_WAITER.endAsync(generator);
+        }
+        return val;
+      } catch (e) {
+        GENERATOR_WAITER.endAsync(generator);
```
rwjblue, line 162: Lets use `finally` instead of `catch` here (no need to re-throw, and the devtools "stop on uncaught exceptions" will still work)

addon/@ember/test-waiters/wait-for.ts
```diff
@@ -96,3 +135,33 @@ function isThenable(
...
+        throw e;
+      }
+    },
+  } as CoroutineGenerator<T>;
+}
```
rwjblue, line 167: This seems vaguely wrong to my eyes (though maybe the coffee just hasn't kicked in yet). We basically need to follow these semantics:

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Iteration_protocols#The_iterator_protocol

As far as I can tell we have a few problems with the current implementation:

* we cannot return `val` from `next` (we have to **always** return our custom iterable, or we don't catch subsequent async in follow up `next` calls)
* We should bracketing the start/stop of async around calling the original iterables `next` (we can't start once before the whole thing, then end after the first next)
* our return value does not have `done` (but AFAICT it must)

addon/@ember/test-waiters/wait-for.ts
```diff
@@ -96,3 +135,33 @@ function isThenable(
     typeof maybePromise.then === 'function'
   );
 }
+
+function isGenerator(
+  maybeGenerator: any | CoroutineGenerator<unknown>
+): maybeGenerator is CoroutineGenerator<unknown> {
+  return typeof maybeGenerator[Symbol.iterator] === 'function';
```
rwjblue, line 142: We can't rely on `Symbol` being defined (we still support IE11 without polyfills).

addon/@ember/test-waiters/wait-for.ts
```diff
@@ -80,6 +117,8 @@ function wrapFunction(fn: Function, label?: string) {
 
     if (isThenable(result)) {
       return waitForPromise(result, label);
+    } else if (isGenerator(result)) {
```
rwjblue, line 120: I think technically you don't want `isGenerator` check here, you want to check if `result` conforms to the iterable protocol (does it have a `done` and a `next` function).

addon/@ember/test-waiters/wait-for.ts
```diff
@@ -96,3 +135,33 @@ function isThenable(
...
+
+function isGenerator(
+  maybeGenerator: any | CoroutineGenerator<unknown>
+): maybeGenerator is CoroutineGenerator<unknown> {
+  return typeof maybeGenerator[Symbol.iterator] === 'function';
+}
+
+const GENERATOR_WAITER = buildWaiter('@ember/test-waiters:generator-waiter');
+
+function waitForGenerator<T>(
```
rwjblue, line 147: Lets rename this to `waitForIterator` (generator is the function which returns an iterator, right?)...

addon/@ember/test-waiters/wait-for.ts
bendemboski: Either you're missing something or I am 😆 I think the naming is confusing -- a [Generator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Generator) (object) *is* an iterator that is returned by a [Generator Function](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/function*). And it's also an iterable. `ember-concurrency` only uses the iterator interface not the iterable interface, which I think might be the source of your confusion?

When we called the function that we're wrapping [here](https://github.com/emberjs/ember-test-waiters/pull/173/files#diff-41ec033b81ca85a4a578fb8082225f03R116), if it was a generator function, we got back a generator object, so that's what we're dealing with here. We're proxying the iterator (not iterable) protocol, which only consists of a `next()` method, so we're returning an object that also implements the iterator protocol, and whose `next()` method just forwards to the underlying/wrapped generator object's `next()` method, but watches for its return value indicating that the iteration/generation is done, or for it throwing an error, so it can stop the waiter. So the ordering here is:

1. We call the generator function to get back a generator object/iterator
2. We start test-waiting
3. We return a wrapper object implementing the iterator procotol
4. Sometime later, someone calls `next()` on our wrapper object, we call `next()` on the generator object, it returns a result saying it's not done iterating, and we return that result
5. One of:
   * Sometime later, someone calls `next()` again on our wrapper object, we call `next()` on the generator object, it returns a result saying it is done iterating, we stop test-waiting and return the result
   * Sometime later, someone calls `next()` again on our wrapper object, we call `next()` on the generator, it throws an error, we stop test-waiting and re-throw the error

I'm pretty sure this is all nominally correct, although I've realized a few things from being challenged (the internet is hard, so just want to emphasize that in this context I see being challenged as a 100% good thing!) and thinking through it again:

1. Generator objects also implement `return()` and `throw()`. I think `ember-concurrency` uses `throw()` for task cancellation? Anyway, we should definitely proxy those methods onto our wrapper object as well.
2. While I don't think `ember-concurrency` does this, callers can continue calling `next()` on generator objects after the iteration has completed, and they'll continue getting `{ value: undefined, done: true }`. In this case we would continue calling `endAsync()` on the waiter, which is very bad, so I suppose we probably need an `isDone` function-scoped variable to track when the generator has run to completion and we've already ended test-waiting so we don't do it again.
3. A generator object's `next()` accepts an argument that shows up in the generator function as a return value from the previous `yield` statement -- I'm pretty certain `ember-concurrency` never does this, but I suppose we should forward arguments from our wrapper's `next()` function to the generator object's `next()` function anyway
4. Since generator objects also implement the iterable protocol, we could implement `[Symbol.iterator]` on our proxy object and have it just return itself (the proxy object).
5. We should definitely write some tests covering basic `ember-concurrency` functionality (multi-`yield` task functions, throwing errors from the task function, task cancellation, etc.) to make sure our wrapping doesn't interfere with anything.
6. We should *maybe* write some generator object interface tests to unit test that our wrapper object doesn't interfere with the generator object interface? I'm not really sure how far we need to go with this right now since currently our only known use case is `ember-concurrency`.

Some of these definitely make sense to do now. But it's not clear to me how much effort it makes sense to spend right now on cases that don't happen in `ember-concurrency`.

Let me know your thoughts.

addon/@ember/test-waiters/wait-for.ts
bendemboski: See big comment below -- pretty sure we need this catch block since this try...catch happens once per iteration/`yield` statement, not once per full-iteration-over-the-generator-function, i.e. (in `ember-concurrency` terms) once per yielded promise, not once per task execution.

addon/@ember/test-waiters/wait-for.ts
bendemboski: How do we handle this then? I guess use duck typing to check for the generator interface, then? Check for `next`, `throw`, and `return` methods? Technically I suppose we could just check for a `next()` method, but I think I feel better about targeting this to generator functions (well, functions that return something that implements the generator object interface) rather than any function that returns something implementing the iterator interface.

bendemboski: Related (from my big comment above) if we do proxy the generator object's `@@iterator` method in `waitForGenerator()`, how would we handle that in IE11? Do you know offhand, or should I fire up my Windows VM and look at the result of a transpiled generator function invocation?

addon/@ember/test-waiters/wait-for.ts
bendemboski: As stated above, I think we do want to look for a generator object here. We could potentially relax it to look for anything implementing the iterator protocol (which I'm pretty sure only has a `next` method), but as stated (further) above, I'm a little uncomfortable with being that loose when we don't know of any uses cases that require us to handle iterators-but-not-generators.

addon/@ember/test-waiters/wait-for.ts
bendemboski: I believe this naming is accurate -- see above

bendemboski:
@rwjblue I've made the fixes and improvements I outlined in https://github.com/emberjs/ember-test-waiters/pull/173#discussion_r467480427 (and task cancellation was in fact not working) and also got rid of the `Symbol.iterator` reference. Let me know if you have any other thoughts on the comments above, or any other feedback -- I'm considering the ball as being out of my court now.

addon/@ember/test-waiters/wait-for.ts
scalvert: Hmmm I think I answered my own question. The generator function returns a generator object. 👍

addon/@ember/test-waiters/wait-for.ts
```diff
@@ -1,18 +1,25 @@
 import { DEBUG } from '@glimmer/env';
 import waitForPromise from './wait-for-promise';
+import buildWaiter from './build-waiter';
 import { PromiseType } from './types';
 
+type CoroutineGenerator<T> = Generator<any, T, any>;
 type AsyncFunction<A extends Array<any>, PromiseReturn> = (...args: A) => Promise<PromiseReturn>;
-type AsyncFunctionArguments = [AsyncFunction<any[], any>, string?];
+type TaskFunction<A extends any[], T> = (...args: A) => CoroutineGenerator<T>;
```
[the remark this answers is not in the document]
bendemboski: No, we're not specifically thinking of/limiting to `ember-concurrency` here. I was thinking of `task` in the completely generic sense of a thing you're trying to do. That's also the language `ember-concurrency` uses, but it obviously didn't invent that concept, so I wasn't thinking of using `task` as explicitly referring to `ember-concurrency` -- just to a concept that it also uses. That being said, I chose it with fairly minimal thought, so am definitely open to different naming. `GeneratorFunction` is already a type (an interface), so I don't think that's a good one to use -- I suppose `CoroutineGeneratorFunction` would make sense since it's consistent with `CoroutineGenerator`.

addon/@ember/test-waiters/wait-for.ts
```diff
@@ -38,21 +45,51 @@ type DecoratorArguments = [object, string, PropertyDescriptor, string?];
...
-  ...args: AsyncFunctionArguments | DecoratorArguments
-): PropertyDescriptor | Function {
-  let isAsyncFunction = args.length < 3;
+  ...args: FunctionArguments | DecoratorArguments
+): PropertyDescriptor | Function | TaskFunction<any[], any> {
+  let isFunction = args.length < 3;
 
-  if (isAsyncFunction) {
-    let [fn, label] = args as AsyncFunctionArguments;
+  if (isFunction) {
```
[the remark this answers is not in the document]
bendemboski: I disagree. I would argue that `isAsyncFunction` was actually a misnomer. `args.length < 3` tells us nothing about what kind of thing this function is wrapping, it only tells us how this function has been invoked -- as a function call or as a decorator.

At this point we need to make the decision of *how* to apply `wrapFunction()` to wrap the thing the caller wants us to wrap, not *what* the thing is the caller wants us to wrap. It's the function that `wrapFunction()` generates that determines *what* the thing is because we can't possibly distinguish between an async function, a non-async function, and a generator function until we actually call it and inspect the return value, which we can't do here.

So at this point we can really only have two cases, wait-for-was-invoked-as-a-function-call and wait-for-was-invoked-as-a-decorator, but if you like, I could clarify by renaming it to `isFunctionalInvocation` or something like that.

addon/@ember/test-waiters/wait-for.ts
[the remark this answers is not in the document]
bendemboski: 👍

addon/@ember/test-waiters/wait-for.ts
```diff
@@ -1,18 +1,25 @@
 import { DEBUG } from '@glimmer/env';
 import waitForPromise from './wait-for-promise';
+import buildWaiter from './build-waiter';
 import { PromiseType } from './types';
 
+type CoroutineGenerator<T> = Generator<any, T, any>;
 type AsyncFunction<A extends Array<any>, PromiseReturn> = (...args: A) => Promise<PromiseReturn>;
-type AsyncFunctionArguments = [AsyncFunction<any[], any>, string?];
+type TaskFunction<A extends any[], T> = (...args: A) => CoroutineGenerator<T>;
+type FunctionArguments = [AsyncFunction<any[], any> | TaskFunction<any[], any>, string?];
```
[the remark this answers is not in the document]
bendemboski: 👍

addon/@ember/test-waiters/wait-for.ts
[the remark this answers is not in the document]
scalvert: Ya I think my relevant point is that in the prior implementation there were only two distinguishable paths: async functions and decorators. With this addition it's true that inspecting the length of the arguments isn't enough to unequivocally determine this, but my point still stands. I think the current version is less readable, and as such may be less maintainable. Is there anything we can do to address this?

addon/@ember/test-waiters/wait-for.ts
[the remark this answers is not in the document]
scalvert: Ya I personally feel like if we're sticking to a taxonomy for our internal branches, we should keep them consistent. I'm 👍 for the names you suggested.

addon/@ember/test-waiters/wait-for.ts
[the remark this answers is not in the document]
bendemboski: No, there's nothing we can do. In this function, the only way to be able to distinguish three paths instead of two would be if we could distinguish a generator function from an async function without calling that function, and that's not possible. Moreover, I will argue again that it's not now and never was thins function's job to do that.

`async functions and decorators` were not the two cases in the prior implementation. There were actually four cases:

1. Called in functional form, passed an async function
2. Called in function form, passes a "sync" function
3. Called in decorator form, passed an async function
4. Called in decorator form, passed a "sync" function

`async function` and `decorator` are not equivalent paths -- `called as a function` and `called as a decorator` are equivalent paths, and `called on an async function` and `called on a regular function` (and now `called on a generator function`) are equivalent paths. This function's job is to distinguish between `called as a function` and `called as a decorator` so it knows *how* to wrap the function it was passed, and the wrapper function returned from `wrapFunction()` is the one whose job it is to distinguish between `called on an async function` and `called on a regular function` (and now `called on a generator function`).

That's why I say `isAsyncFunction` was a misnomer -- it was incorrectly conflating how `waitFor()` was invoked (function or decorator) with what was passed to `waitFor()` (async function or regular function).

Make sense?

bendemboski:
@rwjblue @scalvert bump!

bendemboski:
@rwjblue @scalvert this is starting to get really frustrating. This PR has been stalled for almost two months, and it's been almost 11 months since the origin `waitForCoroutine` PR was opened. In the meantime I'm continuing to have to maintain and update `ember-concurrency-test-waiter` which I really want to deprecate in favor of `waitForCoroutine`, and I'm sure it's only a matter of time until more significant changes are made to `ember-test-waiters` requiring additional effort to rebase this PR and resolve conflicts.

Should I move this implementation to a separate library so I can roll it out and finally deprecate `ember-concurrency-test-waiter`? Or can we drive this to completion?

addon/@ember/test-waiters/wait-for.ts
scalvert: This makes sense, @bendemboski. I'm a bit concerned about calling `endAsync` here mid-iteration, as it may open the testing system up to unpause waiting and incorrectly signal `settled`. WDYT?

bendemboski: I think we're in the uncanny valley between thinking of this in terms of general iterator semantics and thinking of it in terms of a coroutine/ember-concurrency task.

I suppose in terms of general iterator semantics, the user might catch this exception but not discard the iterator, then fix some state or something, and then continue iteration.

But in the context of coroutines and `ember-concurrency`, this means that the coroutine/task has terminated by throwing an exception, and execution will not resume. This is the equivalent of an exception propagating out of an async function, causing the promise it returned to reject, and irreversibly terminating the execution of the function. So this is really the equivalent of [this](https://github.com/emberjs/ember-test-waiters/blob/cccdb301f3/addon/%40ember/test-waiters/wait-for-promise.ts#L44), and the fact that the user could theoretically resume iteration is just a side-effect of the fact that we're using iterator semantics to implement async behavior.

bendemboski: So, we are making the assumption here that the wrapping/management code interprets a thrown exception as "terminate execution," but in all of the contexts that we can think of this being used, that's accurate, and I don't think we can possibly do without that assumption (because then we have no idea if/when the user will resume iteration, so we might never have another opportunity `endAsync`, causing `settled()` to hang forever)

scalvert:
@bendemboski I can understand the frustration. Sorry this has taken so long. @rwjblue and I have been trying to manage a succession of other test library updates, and this one fell by the wayside.

We discussed this PR today and will try our best to move it forward as quickly as possible.

README.md
```diff
@@ -157,6 +157,34 @@ export default class MoreFriendz extends Component {
...
+
+```js
+// decorator form with ember-concurrency tasks
+import Component from '@ember/component';
+import { task } from 'ember-concurrency-decorators';
+import { waitFor } from 'ember-test-waiters';
+
+export default class Friendz extends Component {
+  @task
+  @waitFor
```
scalvert, line 181: Can we add a comment that helps explain the relevance of ordering WRT these decorators?

addon/@ember/test-waiters/wait-for.ts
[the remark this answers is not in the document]
scalvert: Yep, makes perfect sense.

addon/@ember/test-waiters/wait-for.ts
scalvert: This looks correct. We decided to keep this as is.

addon/@ember/test-waiters/wait-for.ts
scalvert: We agreed this is correct, as we're specifically targeting generators.

addon/@ember/test-waiters/wait-for.ts
```diff
@@ -96,3 +138,54 @@ function isThenable(
...
+    if (isWaiting) {
+      GENERATOR_WAITER.endAsync(generator);
+      isWaiting = false;
+    }
+  }
+
+  return {
+    next(...args) {
+      try {
+        let val = generator.next(...args);
```
rwjblue, line 172: I _vaguely_ think we need to rewrap in this case, right?

addon/@ember/test-waiters/wait-for.ts
```diff
@@ -96,3 +138,54 @@ function isThenable(
...
+        if (val.done) {
+          stopWaiting();
+        }
+        return val;
+      } catch (e) {
```
rwjblue, line 177: I'd like to avoid `try`/`catch` here. Can you swap to using `try`/`finally` instead?

addon/@ember/test-waiters/wait-for.ts
rwjblue: We chatted through this and since we are only doing Generators (not the more general iterator) this is fine.

scalvert (changes requested):

scalvert:
Final things to add:

- [ ] Document the decorator ordering
- [ ] Migrating away from try/catch (and using @rwjblue's reverso boolean check)
- [ ] Add test for `ember-concurrency` nested/child task cancellation 
- [ ] Optional, if not too hard, add an additional generator implementation to test against (such as `co`)

bendemboski:
Okay, I think this is all set to go (build failing because #195)

rwjblue:
Ready for a rebase now that #195 landed.

addon/@ember/test-waiters/wait-for.ts
```diff
@@ -96,3 +138,63 @@ function isThenable(
...
+        if (val.done) {
+          stopWaiting();
+        }
+        return val;
+      } finally {
+        // If generator.next() throws, we need to stop waiting. But if we catch
+        // and re-throw exceptions, it could move the location from which the
+        // uncaught exception is thrown, interfering with the developer
+        // debugging experience if they have break-on-exceptions enabled. So we
+        // use a boolean flag and a finally block to emulate a catch block.
```
rwjblue, line 185: For those that are curious, here is a smallish codepen that demos the dev ergonomics issue I was referring to: https://codepen.io/rwjblue/full/PozYKqw.

When you run it, make sure to have DevTools open and "Pause on uncaught exceptions" set in the sources tab. The execution **should** pause in the `someBadFunction` implementation where `throw new Error()` is called, but it is actually pausing inside the `catch` block where we `throw e`.

rwjblue (approved):

rwjblue:
2.18 failure is:

```
not ok 45 Chrome 86.0 - [7 ms] - wait-for > waitFor ember-concurrency interop: tasks with multiple yields work
    ---
        actual: >
            null
        stack: >
            Error: Assertion Failed: [object Object] does not appear to be a task!
                at new EmberError (http://localhost:7357/assets/vendor.js:25479:25)
                at Object.assert (http://localhost:7357/assets/vendor.js:25722:15)
                at perform (http://localhost:7357/assets/vendor.js:67990:69)
                at _callee8$ (http://localhost:7357/assets/tests.js:1258:63)
                at tryCatch (http://localhost:7357/assets/vendor.js:413:40)
                at Generator.invoke [as _invoke] (http://localhost:7357/assets/vendor.js:687:22)
                at Generator.prototype.<computed> [as next] (http://localhost:7357/assets/vendor.js:446:21)
                at asyncGeneratorStep (http://localhost:7357/assets/tests.js:644:105)
                at _next (http://localhost:7357/assets/tests.js:646:196)
                at http://localhost:7357/assets/tests.js:646:366
        message: >
            Promise rejected during "tasks with multiple yields work": Assertion Failed: [object Object] does not appear to be a task!
        negative: >
            false
        browser log: |
    ...
not ok 46 Chrome 86.0 - [5 ms] - wait-for > waitFor ember-concurrency interop: direct task cancellation works
    ---
        actual: >
            null
        stack: >
            Error: Assertion Failed: [object Object] does not appear to be a task!
                at new EmberError (http://localhost:7357/assets/vendor.js:25479:25)
                at Object.assert (http://localhost:7357/assets/vendor.js:25722:15)
                at perform (http://localhost:7357/assets/vendor.js:67990:69)
                at _callee9$ (http://localhost:7357/assets/tests.js:1315:66)
                at tryCatch (http://localhost:7357/assets/vendor.js:413:40)
                at Generator.invoke [as _invoke] (http://localhost:7357/assets/vendor.js:687:22)
                at Generator.prototype.<computed> [as next] (http://localhost:7357/assets/vendor.js:446:21)
                at asyncGeneratorStep (http://localhost:7357/assets/tests.js:644:105)
                at _next (http://localhost:7357/assets/tests.js:646:196)
                at http://localhost:7357/assets/tests.js:646:366
        message: >
            Promise rejected during "direct task cancellation works": Assertion Failed: [object Object] does not appear to be a task!
        negative: >
            false
        browser log: |
    ...
not ok 47 Chrome 86.0 - [14 ms] - wait-for > waitFor ember-concurrency interop: parent task cancellation works
    ---
        actual: >
            null
        stack: >
            Error: Assertion Failed: [object Object] does not appear to be a task!
                at new EmberError (http://localhost:7357/assets/vendor.js:25479:25)
                at Object.assert (http://localhost:7357/assets/vendor.js:25722:15)
                at perform (http://localhost:7357/assets/vendor.js:67990:69)
                at _callee9$ (http://localhost:7357/assets/tests.js:1315:66)
                at tryCatch (http://localhost:7357/assets/vendor.js:413:40)
                at Generator.invoke [as _invoke] (http://localhost:7357/assets/vendor.js:687:22)
                at Generator.prototype.<computed> [as next] (http://localhost:7357/assets/vendor.js:446:21)
                at asyncGeneratorStep (http://localhost:7357/assets/tests.js:644:105)
                at _next (http://localhost:7357/assets/tests.js:646:196)
                at http://localhost:7357/assets/tests.js:646:366
        message: >
            Promise rejected during "parent task cancellation works": Assertion Failed: [object Object] does not appear to be a task!
        negative: >
            false
        browser log: |
    ...
not ok 48 Chrome 86.0 - [11 ms] - wait-for > waitFor ember-concurrency interop: wrapped parent task cancellation works
    ---
        actual: >
            null
        stack: >
            Error: Assertion Failed: [object Object] does not appear to be a task!
                at new EmberError (http://localhost:7357/assets/vendor.js:25479:25)
                at Object.assert (http://localhost:7357/assets/vendor.js:25722:15)
                at perform (http://localhost:7357/assets/vendor.js:67990:69)
                at _callee9$ (http://localhost:7357/assets/tests.js:1315:66)
                at tryCatch (http://localhost:7357/assets/vendor.js:413:40)
                at Generator.invoke [as _invoke] (http://localhost:7357/assets/vendor.js:687:22)
                at Generator.prototype.<computed> [as next] (http://localhost:7357/assets/vendor.js:446:21)
                at asyncGeneratorStep (http://localhost:7357/assets/tests.js:644:105)
                at _next (http://localhost:7357/assets/tests.js:646:196)
                at http://localhost:7357/assets/tests.js:646:366
        message: >
            Promise rejected during "wrapped parent task cancellation works": Assertion Failed: [object Object] does not appear to be a task!
        negative: >
            false
        browser log: |
    ...
```

bendemboski:
Ack! Trolled by Ember.get() (well, the lack thereof)!

bendemboski:
Okay, all sorted!

scalvert (approved):

scalvert:
Thanks, @bendemboski!

PR #174, opened 2020-08-12
docs: fix install instructions to use proper pkg name

jamescdavis:
The installation instructions in the readme use the module name (`@ember/test-waiters`), but the package is published as `ember-test-waiters`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,7 +52,7 @@ This addon implements the design specified in [RFC 581](https://github.com/ember
 ## Installation
 
 ```
-ember install @ember/test-waiters
+ember install ember-test-waiters
 ```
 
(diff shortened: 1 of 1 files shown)


rwjblue:
> The installation instructions in the readme use the module name (@ember/test-waiters), but the package is published as ember-test-waiters.

Sorry about the confusion here @jamescdavis, we were actually in the middle of a final push towards the new package name and let the docs changes land a bit too much before the actual publishing. Sorry about that!

Closing, as `@ember/test-waiters` **is** the correct name (as of ~ 6 hours ago):

https://www.npmjs.com/package/@ember/test-waiters

PR #195, opened 2020-10-27
Fix CI build

bendemboski:
Work around https://github.com/actions/virtual-environments/issues/1811


rwjblue (approved):
Seems fine to me, nothing we are doing should really platform dependent.

PR #231, opened 2021-02-12
update `cacheKeyForTree` override to account for both addon names

brendenpalmer:
This updates the `cacheKeyForTree` override to account for both addon names (ie, `@ember/test-waiters` and `ember-test-waiters`).

The reason for this is that Embroider doesn't currently take into account the addon's name when dealing with cache keys internally & as a result, it gets confused that (seemingly) two different addons have the same cache key. This will eventually be fixed in Embroider, but in the short-term we're implementing this here to account for this issue.

The end result is that we get a different stable cache key, one for `@ember/test-waiters` and another for `ember-test-waiters`.

### DIFF
--- a/force-highlander-addon.js
+++ b/force-highlander-addon.js
@@ -1,6 +1,7 @@
 'use strict';
 const semver = require('semver');
 const VersionChecker = require('ember-cli-version-checker');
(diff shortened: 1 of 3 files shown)


rwjblue:
Thank you @brendenpalmer!

PR #232, opened 2021-02-24
Fix potentially massive memory leak

bendemboski:
The logic for choosing between the Map and WeakMap for completed operation tracking wasn't right, and the result was that everything was being added to the Map, leaking every token. For waitForPromise() the token is the promise itself, meaning the resolved values were also leaking, which can easily add up to a memory leak of epic proportions.

### DIFF
--- a/addon/@ember/test-waiters/build-waiter.ts
+++ b/addon/@ember/test-waiters/build-waiter.ts
@@ -89,9 +89,11 @@ class TestWaiterImpl<T extends object | Primitive = Token> implements TestWaiter
   private _getCompletedOperations(token: T) {
     let type = typeof token;
 
-    return token !== null || (type !== 'function' && type !== 'object')
-      ? this.completedOperationsForPrimitives
-      : this.completedOperationsForTokens;
+    let isFunction = type === 'function';
+    let isObject = token !== null && type === 'object';
+    let isPrimitive = !isFunction && !isObject;
+
+    return isPrimitive ? this.completedOperationsForPrimitives : this.completedOperationsForTokens;
   }
 }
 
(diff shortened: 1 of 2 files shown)


bendemboski:
I guess `waitFor()` on a coroutine is also extremely leaky since the token is the generator function itself, so its context probably includes `this`, meaning it will leak the containing object.

scalvert:
Woah, thanks for finding this, @bendemboski. I'll take a look.

scalvert:
Derp, so we'd essentially reversed the logic for token storage. Nice catch!

scalvert (approved):

addon/@ember/test-waiters/build-waiter.ts
```diff
@@ -89,7 +89,7 @@ class TestWaiterImpl<T extends object | Primitive = Token> implements TestWaiter
   private _getCompletedOperations(token: T) {
     let type = typeof token;
 
-    return token !== null || (type !== 'function' && type !== 'object')
+    return token === null || (type !== 'function' && type !== 'object')
```
rwjblue, line 92: Can you make this mistake harder to make?

I'm thinking something like:

```suggestion
    let isFunction = type === 'function';
    let isObject = token !== null && type === 'object';

    return isFunction || isObject
```

bendemboski: Sure, how about

```js
    let isFunction = type === 'function';
    let isObject = token !== null && type === 'object';
    let isPrimitive = !isFunction && !isObject;

    return isPrimitive ? this.completedOperationsForPrimitives : this.completedOperationsForTokens;
```

rwjblue: Ya, seems great to me!

scalvert: I like.

rwjblue (approved):

PR #239, opened 2021-03-11
Broken type imports leak into Javascript

ef4:
The meat of this PR is the one line change in https://github.com/emberjs/ember-test-waiters/commit/31317db4c583977b63e4f5b0b1cf0777498a8f89.

To make that work though I had to upgrade prettier, which required me to reformat a lot of the repo, and upgrade ember-cli-typescript and ember-cli-babel to make sure we would get a new enough version of `@babel/plugin-transform-typescript` to handle type-only exports.

The motivation for this change is that without marking these are type-only exports, the export statement ends up staying in the Javascript even though the corresponding type definitions got deleted. That mismatch is a SyntaxError when you're using a system that respects ES modules.

### DIFF
--- a/addon/@ember/test-waiters/build-waiter.ts
+++ b/addon/@ember/test-waiters/build-waiter.ts
@@ -68,7 +68,7 @@ class TestWaiterImpl<T extends object | Primitive = Token> implements TestWaiter
   debugInfo(): TestWaiterDebugInfo[] {
     let result: TestWaiterDebugInfo[] = [];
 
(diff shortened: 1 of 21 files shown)


rwjblue (approved):
LGTM - I'll try to land and release in the AM

rwjblue:
released in https://github.com/emberjs/ember-test-waiters/releases/tag/v2.4.2

ef4:
Thanks!

PR #272, opened 2021-06-11
Fix warnings when the legacy module deprecation is triggered

sandydoo:
```
DEPRECATION: When calling `deprecate` you must provide `for` in options. Missing options.for in \"ember-test-waiters-legacy-module-name\" deprecation [deprecation id: ember-source.deprecation-without-for]

DEPRECATION: When calling `deprecate` you must provide `since` in options. Missing options.since in \"ember-test-waiters-legacy-module-name\" deprecation [deprecation id: ember-source.deprecation-without-since]
```

This adds the required `for` and `since` options when creating the deprecation warning.

PR #279, opened 2021-07-12
Adds type checking to CI

scalvert:
Our current CI doesn't do any type checking, which incorrectly signals that everything is valid up until you try to release. During release, we run a build via `ember ts:precompile`, which will error if there are issues with types.

Adding this type check during CI will help avoid this situation.

Relates to #272, depends on #280


scalvert:
CI will fail since we currently have a type mismatch due to #272

.github/workflows/ci-build.yml
```diff
@@ -23,6 +23,8 @@ jobs:
         run: yarn install
       - name: lint
         run: yarn lint:js
+      - name: validate
+        run: yarn tsc --noEmit
```
rwjblue, line 27: We already have a script for this `yarn lint:ts`

https://github.com/emberjs/ember-test-waiters/blob/1f4ed4d96c4e5958c4ab286f734ee052fd60e236/package.json#L20-L31

We should tweak to match the more modern addon blueprint for our scripts, so we run `yarn lint` in CI and that runs `lint:*` (along with `yarn test` running `yarn lint`, to ensure that local testers just running `yarn test` have a proper "pass"/"fail" experience).

rwjblue (changes requested):

.github/workflows/ci-build.yml
scalvert: Oh good call. I'll update to that.

scalvert: OK, I've updated the scripts to more closely match the blueprint, though there are some small differences. Let me know if you think that makes sense.

rwjblue (approved):

PR #280, opened 2021-07-12
Fixes TS issue with mismatched type for deprecate

scalvert:
Fixes type mismatch introduced in #272.

### DIFF
--- a/addon/ember-test-waiters/index.ts
+++ b/addon/ember-test-waiters/index.ts
@@ -7,7 +7,9 @@ deprecate(
     id: 'ember-test-waiters-legacy-module-name',
     until: '3.0.0',
     for: 'ember-test-waiters',
-    since: '2.2.0',
+    since: {
+      enabled: '2.2.0',
+    },
   }
 );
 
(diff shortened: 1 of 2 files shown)


NullVoxPopuli:
should an issue be opened to ember-cli-typescript as well?

scalvert:
What specifically do you think that issue should call out?

NullVoxPopuli:
To update https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/ember__debug/index.d.ts#L71 to match the current API so that test-waiters can eventually remove the override?

scalvert:
Oh, so that would be in DefinitelyTyped, not `ember-cli-typescript`, right? Also, I believe there's already a PR for this: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/52397

scalvert:
I've also further updated the call site for `deprecate` since the `since` property was incorrectly called (it's an object, not a string).

rwjblue (approved):

PR #293, opened 2021-08-12
Updates publish step to publish JavaScript + *.d.ts vs. .ts files.

### LINKED ISSUE #290: TS Files appear to be published
It appears this projects TS files are being publish, I am under the impression that we want to publish the build `js` + `d.ts` but not the `ts`. Has this change or...

Reproduction:
```sh
stefanpenner@stefan /t/foo [23|1]> curl (npm info @ember/test-waiters --json | jq -r '.dist.tarball') | tar xvz -
x package/.prettierrc.js
x package/config/environment.js
x package/node-tests/force-highlander-addon-test.js
x package/force-highlander-addon.js
x package/index.js
x package/vendor/shims/qunit.js
x package/package.json
x package/tsconfig.json
x package/CHANGELOG.md
x package/LICENSE.md
x package/README.md
x package/RELEASE.md
x package/build-waiter.d.ts
x package/addon/@ember/test-waiters/build-waiter.ts
x package/types/custom-ember-debug.d.ts
x package/types/global.d.ts
x package/index.d.ts
x package/types/dummy/index.d.ts
x package/types/index.d.ts
x package/types/qunit/index.d.ts
x package/addon/@ember/test-waiters/index.ts
x package/addon/@ember/test-waiters/types/index.ts
x package/addon/ember-test-waiters/index.ts
x package/token.d.ts
x package/addon/@ember/test-waiters/token.ts
x package/wait-for-promise.d.ts
x package/addon/@ember/test-waiters/wait-for-promise.ts
x package/wait-for.d.ts
x package/addon/@ember/test-waiters/wait-for.ts
x package/waiter-manager.d.ts
x package/addon/@ember/test-waiters/waiter-manager.ts
x package/.github/workflows/ci-build.yml
x package/.github/dependabot.yml
```

stefanpenner:
Fixes #290 

This moves the ember-cli-typescript to a dev dependency, and does not require the consuming app know, or care how to consume typescript.

I've left a more detailed discussion issue over at ember-cli-typescript https://github.com/typed-ember/ember-cli-typescript/issues/1443


package.json
```diff
@@ -24,7 +24,11 @@
     "lint:js": "eslint --ext ts,js .",
     "lint:js:fix": "eslint --ext ts,js . --fix",
     "lint:ts": "tsc -p tsconfig.json --noEmit",
-    "prepublishOnly": "ember ts:precompile",
+
+    "prepublishOnly": "npm-run-all prepublishOnly:*",
+    "prepublishOnly:d.ts": "ember ts:precompile",
+    "prepublishOnly:addon": "babel --extensions '.ts' --presets @babel/preset-typescript addon --out-dir addon --ignore '**/*.d.ts'",
```
stefanpenner, line 30: Ultimately, I think `ember test:precompile" should do this for us

scalvert:
Let's wait for a discussion on this. I'm fine to make changes in this space, but I want to do what the standard is rather than a one-off.

stefanpenner:
@scalvert and me chatted, he is ok with a few libraries (such as this one) being part of some early experimentation in the space as we build up some evidence + examples to properly improve both ember-cli-typescript and embroiders v2 package format with regards to ts support.

This is also important, as without this several consumers are broken, which implies a breaking change has already landed.

I do agree a good standard here will be important, and we can use this short-term fix as information to help solve that bigger picture. Once solved we can loop back and upgrade this project

package.json
```diff
@@ -24,7 +24,9 @@
     "lint:js": "eslint --ext ts,js .",
     "lint:js:fix": "eslint --ext ts,js . --fix",
     "lint:ts": "tsc -p tsconfig.json --noEmit",
-    "prepack": "ember ts:precompile",
+    "prepack": "npm-run-all prepublishOnly:*",
```
scalvert, line 27: Shouldn't this be:

```suggestion
    "prepack": "npm-run-all prepack:*",
```

?

scalvert (approved):

chriskrycho:
@stefanpenner can you elaborate on this?

> This is also important, as without this several consumers are broken, which implies a breaking change has already landed.

It's not clear to me *how* things were broken, and without that it's difficult to understand why this is the solution. Per discussion on typed-ember/ember-cli-typescript#1443, we want to move to align the Ember ecosystem with the rest of the TS ecosystem as part of the move to Embroider, but the way this worked historically has actually never required consuming apps or addons to know or care that there's TS involved: it happens transparently for them. If it doesn't, that means that (a) something is wrong in the publishing package; or (b) that there's a bug somewhere in the ember-cli-typescript + ember-cli-babel chain.

While (b) is definitely possible, a very large (and ever-growing!) number of packages in the ecosystem work just fine with the current flow, so it would be helpful to know *exactly* what you saw here.

(Also, as a point of clarification: there's basically zero extra overhead for consumers in the way we currently publish addons, because it's just another Babel transform. The *only* time that type checking gets involved is if the consuming addon *itself* chooses to add ember-cli-typescript—and even then, it has the choice of whether to type check its dependencies via the `compilerOptions.skipLibCheck` setting in `tsconfig.json`.)

stefanpenner:
@chriskrycho its been to long, but a a semver drift of `@ember/test-helpers` broke a consuming app or add-on due to this issue in this repo.

Unfortunately I've been working my way through too many add-ons the last few months, I wish I could have noted it in this issue.

PR #297, opened 2021-08-12
Fix canary

stefanpenner:
This fixes the canary try-scenario... let's see how the other scenarios appreciate it...

### DIFF
--- a/.github/workflows/ci-build.yml
+++ b/.github/workflows/ci-build.yml
@@ -5,27 +5,41 @@ on:
     branches:
       - master
       - 'v*'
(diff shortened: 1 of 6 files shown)


stefanpenner:
Looks like ember 2.18 isn't happy, I'll poke around and see if there is a happy medium between supporting ember-canary and ember-lts-2.18.

We may want to just drop support for 2.18.... @rwjblue / @scalvert WDYT?

stefanpenner:
Speaking with @scalvert, we should likely match ember-try support with ember-test-helpers

PR #305, opened 2021-08-13
Update ci-build.yml

scalvert:
The timeout added in #297 has caused jobs to time out. This PR removes that.


stefanpenner:
The time-out was added intentionally, as having tests that hang, will currently take ages to timeout naturally clogging up our CI pipeline.

If the timeout duration is in-appropriate, we can adjust the threshold.

PR #306, opened 2021-08-16
Fix Failing Tests resulting from ember-concurrency v2 bump

stefanpenner:
1) treat `continue()` not as an idempotent function, but rather as a once time resource. This helps detect failures, specifically ones where we are cooperating with another co-routine. In-practice, this means `continue()` will throw if invoked twice for the same continuation

2) `continue()` now returns a promise, which tests now await rather than Promise.resolve. This enables the generator itself to dictate when the the active continuation has come to an end.

—

After debugging, it really just came down to the fact that `await Promise.resolve()` and the `yieldable` used in this test implicitly depended on something they really should not have. Ember 3.16 and 3.20 had slightly different glimmer dirty-tracking behaviors, which resulted in the mixture of native promises and ember auto-run setTimeouts creating a situation where the precariousness of the relationship between the above mentioned operations was exposed.

The proper way to address this is to ensure the test code makes that relationship explicit. I suspect ember-concurrencies new yieldable may be the recommended approach, but this should suffice to get the tests green again.

I have also brain dumped all this on poor Steve, who may look into the yieldable stuff.

### DIFF
--- a/tests/unit/wait-for-test.ts
+++ b/tests/unit/wait-for-test.ts
@@ -234,18 +234,54 @@ if (DEBUG) {
 
       class NativeThing {
         iterations: Array<Number> = [];
-        continue: Function = () => null;
+        _continue?: Function;
 
         @taskDec
         @waitFor
         *doStuffTask(): TaskGenerator<string> {
           for (let i = 0; i < 3; i += 1) {
-            let deferred = new Deferred();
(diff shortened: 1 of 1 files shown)


scalvert:
cc/ @bendemboski for visibility into the test issue.

scalvert (approved):

PR #308, opened 2021-08-16
embroider ember-try scenarios

stefanpenner:
This enables embroider-safe ember-try scenario


stefanpenner:
tests pass manually, but I need to fix testem integration still.

```
 ERROR: Uncaught Error: Testem was unable to detect a test framework, please load it before invoking Testem.hookIntoTestFramework at http://localhost:7357/testem.js, line 913
 ```

**Comment by stefanpenner** [comment]
Interesting, tests pass locally fine. I guess I need to dig in....

**Comment by stefanpenner** [comment]
It looks like (in CI only as far as I can tell), we are attempting to register a webpack AMD style module (with only 1 argument) to a loader.js loader.

**Comment by stefanpenner** [comment]
I can reproduce the (1) the failure (2) the test timeout by running

```sh
env CI=true yarn ember try:one embroider-safe
```

... currently debugging

stefanpenner:
The issue was basically: https://github.com/embroider-build/embroider/issues/677#issuecomment-813118275

ember-cli-build.js
```diff
@@ -3,21 +3,22 @@
 const EmberAddon = require('ember-cli/lib/broccoli/ember-addon');
 
 module.exports = function (defaults) {
-  let app = new EmberAddon(defaults, {
-    autoImport: {
```
stefanpenner, line 7: we don't need this, if we include qunit via auto-import

ember-cli-build.js
```diff
@@ -3,21 +3,22 @@
...
-  app.import('node_modules/qunit/qunit/qunit.js', {
-    type: 'test',
-  });
+  let app = new EmberAddon(defaults);
 
   app.import('node_modules/qunit/qunit/qunit.css', {
     type: 'test',
   });
 
-  app.import('vendor/shims/qunit.js', { type: 'test' });
```
stefanpenner, line 20: use auto-import instead

ember-cli-build.js
```diff
@@ -3,21 +3,22 @@
...
-
-  return app.toTree();
+  try {
```
stefanpenner, line 12: This is verbose, because this library cannot yet drop node 10 and instead requires `@embroider/test-support` be provided via the ember-try-scenario configuration

package.json
```diff
@@ -85,6 +84,7 @@
     "loader.js": "^4.7.0",
     "npm-run-all": "^4.1.5",
     "prettier": "^2.2.1",
+    "promise-polyfill": "^8.2.0",
```
stefanpenner, line 87: 1) now matches `@ember/test-helpers` polyfill
2) works around ie11 + es6-promise + babel + webpack "glitch" that causes some fun to debug issues.

tests/dummy/vendor/ember-cli/test-support-suffix.js
```diff
@@ -0,0 +1,16 @@
+// eslint-disable-next-line no-unused-vars, @typescript-eslint/no-unused-vars
+/* global runningTests: true */
+
+/*
+  used to determine if the application should be booted immediately when `app-name.js` is evaluated
+  when `runningTests` the `app-name.js` file will **not** import the applications `app/app.js` and
+  call `Application.create(...)` on it. Additionally, applications can opt-out of this behavior by
+  setting `autoRun` to `false` in their `ember-cli-build.js`
+*/
+runningTests = true;
```
stefanpenner, line 10: mimic `@ember/test-helpers`

ember-cli-build.js
scalvert: I think the reason we had this here was that we weren't sure of broad auto-import support at the time of writing. Makes sense if it's better supported now.

package.json
scalvert: 👍

tests/index.html
```diff
@@ -22,13 +22,13 @@
...
-    <div id="ember-testing-container">
-      <div id="ember-testing"></div>
+    <div id="qunit-fixture">
+      <div id="ember-testing-container">
+        <div id="ember-testing"></div>
+      </div>
     </div>
 
-    <script src="/testem.js" integrity=""></script>
+    <script src="/testem.js" integrity="" data-embroider-ignore></script>
```
scalvert, line 31: What does this do?

stefanpenner: This ensures embroider doesn't warn that it can't find this, as this is a dynamic files provided by a middleware which shouldn't and cannot be bundled.

PR #312, opened 2021-08-17
ensure `npm pack` is tidy

stefanpenner:
`npm pack` of the latest npm (7.20.6) for some reason wouldn’t behave…unless I made these changes.

Am I going crazy?

----

Anyways... with this, the `npm pack` output is as follows, which appears alright:

```sh
npm notice 1.1kB  LICENSE.md
npm notice 19.4kB README.md
npm notice 3.5kB  addon/@ember/test-waiters/build-waiter.js
npm notice 307B   addon/@ember/test-waiters/index.js
npm notice 460B   addon/@ember/test-waiters/index.ts
npm notice 106B   addon/@ember/test-waiters/token.js
npm notice 10B    addon/@ember/test-waiters/types/index.js
npm notice 1.1kB  addon/@ember/test-waiters/wait-for-promise.js
npm notice 2.5kB  addon/@ember/test-waiters/wait-for.js
npm notice 2.1kB  addon/@ember/test-waiters/waiter-manager.js
npm notice 322B   addon/ember-test-waiters/index.js
npm notice 801B   build-waiter.d.ts
npm notice 1.9kB  force-highlander-addon.js
npm notice 432B   index.d.ts
npm notice 1.9kB  index.js
npm notice 4.0kB  package.json
npm notice 108B   token.d.ts
npm notice 1.0kB  types/custom-ember-debug.d.ts
npm notice 0B     types/dummy/index.d.ts
npm notice 202B   types/global.d.ts
npm notice 2.8kB  types/index.d.ts
npm notice 287B   types/qunit/index.d.ts
npm notice 739B   wait-for-promise.d.ts
npm notice 2.3kB  wait-for.d.ts
npm notice 1.2kB  waiter-manager.d.ts
```

do we need to publish `types/*` ? That seems somewhat odd..


scalvert:
We should be publishing the d.ts files that we published before, whatever the format/dir structure. The difference now is that we're just swapping .ts for .js, so all else should ideally be the same.

What was the weirdness you were referring to?

stefanpenner:
@scalvert the fact that the `npmignore` required i change `/dist/` to `/dist/**/*` to work...

stefanpenner:
Speaking with @hjdivad, he pointed out that NPM is confused by `**/*.d.ts`, it is worth pointing out `gitignore` files handle this correctly. Suggesting that this is indeed and NPM bug, as NPM points to the `gitignore` documentation when reading about `npmignore`.

The solution is to only use `*.d.ts` which does not trigger this problem.

PR #358, opened 2022-01-12
Moving ember-cli-typescript to dependencies

scalvert:
As indicated by the package, this should be a dependency, not a devDep.

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,6 +38,7 @@
   "dependencies": {
     "calculate-cache-key-for-tree": "^2.0.0",
     "ember-cli-babel": "^7.26.6",
+    "ember-cli-typescript": "^4.2.1",
     "ember-cli-version-checker": "^5.1.2",
     "semver": "^7.3.5"
   },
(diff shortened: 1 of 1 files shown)


rwjblue:
Hmm, this doesn't seem quite right. We intentionally do not publish `.ts` files (so we shouldn't need to include this as a dep)...

scalvert:
Ya I think the message in the console is not nuanced/is misleading:

```
WARNING: `ember-cli-typescript` should be included in your `dependencies`, not `devDependencies`
```

I was trusting that direction, but thinking more it does make sense what you're saying, and this message should be updated to apply in the following way:

- for packages that publish .js files to consumers, devDependencies is fine
- for packages that publish .ts files to consumers, dependencies is required.

scalvert:
Happy to revert in light of this. It's not published yet.

chriskrycho:
Just came across this and: yeah, we'll need to update that in light of v2 addons.

scalvert:
This was reverted in https://github.com/emberjs/ember-test-waiters/pull/360

chriskrycho:
Yep! That’s how I got here actually. 😅

PR #388, opened 2022-04-04
Ensure waiters are discovered even if multiple versions exist

rwjblue:
Currently, we rely heavily on the highlander logic to ensure that we can **only** ever have one version of `@ember/test-waiters` in the build outpu. The reason for this is so that when someone calls `hasPendingWaiters()` we can reliably ensure that _**ALL**_ waiters that have been registered (regardless of version mismatches and whatnot) are found.

Unfortunately, we cannot actually rely on the highlander logic to guarantee this "waiter map" uniqueness in Embroider builds (because the highlander logic is broken by the isolated nature of Embroiders compatibility layer).

This change migrates from a strict reliance on the highlander logic to guarantee the module scoped variable is shared globally to a mechanism of sharing the state on `globalThis` (with fallbacks for various other scenarios). Using this pattern allows us to avoid "caring" if the highlander code actually enforces a single version, while still ensuring that `hasPendingWaiters` will always return the right value.

It is likely that future changes will remove the usage of the highlander logic as of a certain version (that includes this change) and higher.

### DIFF
--- a/addon/@ember/test-waiters/waiter-manager.ts
+++ b/addon/@ember/test-waiters/waiter-manager.ts
@@ -3,7 +3,40 @@ import { PendingWaiterState, Waiter, WaiterName } from './types';
 import Ember from 'ember';
 import { registerWaiter } from '@ember/test';
 
-const WAITERS: Map<WaiterName, Waiter> = new Map<WaiterName, Waiter>();
+type Indexable = Record<any, unknown>;
+
+// this ensures that if @ember/test-waiters exists in multiple places in the
+// build output we will still use a single map of waiters (there really should
+// only be one of them, or else `settled` will not work at all)
+const WAITERS: Map<WaiterName, Waiter> = (function () {
+  const HAS_SYMBOL = typeof Symbol !== 'undefined';
+
+  let symbolName = 'LIFELINE_QUEUED_POLL_TASKS';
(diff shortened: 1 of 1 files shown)


addon/@ember/test-waiters/waiter-manager.ts
```diff
@@ -3,7 +3,40 @@ import { PendingWaiterState, Waiter, WaiterName } from './types';
...
-const WAITERS: Map<WaiterName, Waiter> = new Map<WaiterName, Waiter>();
+type Indexable = Record<any, unknown>;
+
+// this ensures that if @ember/test-waiters exists in multiple places in the
+// build output we will still use a single map of waiters (there really should
+// only be one of them, or else `settled` will not work at all)
+const WAITERS: Map<WaiterName, Waiter> = (function () {
+  const HAS_SYMBOL = typeof Symbol !== 'undefined';
+
+  let symbolName = 'LIFELINE_QUEUED_POLL_TASKS';
```
scalvert, line 14: We should probably rename this symbol.

How about:

```suggestion
  let symbolName = 'TEST_WAITERS';
```

scalvert (changes requested):

addon/@ember/test-waiters/waiter-manager.ts
rwjblue: lolol copy pasta!!!!!

scalvert (approved):

PR #452, opened 2023-05-04
v2 Addon prep: move everything in to a single package monorepo

NullVoxPopuli:

`@ember/test-waiters` specific plan here: https://github.com/emberjs/ember-test-waiters/issues/458

--------------

This PR moves the addon to a single package monorepo to prepare for a good few test apps for ensuring we keep compatibility as we prepare for v2-addon-ness.


.github/workflows/ci-build.yml
```diff
@@ -59,9 +59,10 @@ jobs:
...
 
     steps:
-      - uses: actions/checkout@v3
-      - uses: volta-cli/action@v4
-      - name: install dependencies
-        run: npm install
+      - uses: actions/checkout@v4
+      - uses: wyvox/action-setup-pnpm@v3
+        with:
+          node-version: 18.18.1
```
SergeAstapov, line 65: @NullVoxPopuli looks strange we specify Node version only in this job, is that miss or really necessary?

.gitignore
```diff
@@ -1,30 +1,30 @@
...
+.pnp*
+.sass-cache
+connect.lock
+coverage/
+libpeerconnection.log
+npm-debug.log*
+testem.log
+yarn-error.log
 
 # ember-try
```
SergeAstapov, line 22: seem like all the below lines not needed any more and only copy what we have in `addon/.gitignore`?

addon/config/ember-try.js
```diff
@@ -43,7 +43,7 @@ const embroider = {
 
 module.exports = async function () {
   return {
-    useYarn: true,
+    usePnpm: true,
```
SergeAstapov, line 46: IMO it's fine to bundle, maybe we could do this in separate PR. Which may slow down things considering this was idle for some time already.

addon/package.json
```diff
@@ -0,0 +1,123 @@
...
+    "eslint-config-prettier": "^8.3.0",
+    "eslint-plugin-ember": "^7.8.1",
+    "eslint-plugin-node": "^11.1.0",
+    "eslint-plugin-prettier": "^4.0.0",
+    "loader.js": "^4.7.0",
+    "npm-run-all": "^4.1.5",
+    "prettier": "^2.5.1",
+    "promise-polyfill": "^8.2.0",
+    "qunit": "^2.16.0",
+    "release-it": "^14.14.3",
```
SergeAstapov, line 85: not an action: in a follow up we may want to extract this to root level

package.json
```diff
@@ -1,123 +1,19 @@
...
-    "postpack": "ember ts:clean",
-    "start": "ember serve",
-    "test": "ember test",
-    "test:ember": "ember test",
-    "test:ember-compatibility": "ember try:each",
-    "test:prod": "ember test --environment=production"
+    "lint": "pnpm --filter '*' lint:fix",
+    "start": "concurrently 'npm:start:*' --restart-after 5000 --prefix-colors cyan,white,yellow",
+    "start:addon": "pnpm --filter @ember/test-waiters start -- --no-watch.clearScreen",
+    "start:test-app": "pnpm --filter test-app start",
```
SergeAstapov, line 9: technically this is not yet ready, I believe we can keep as it will appear in next PR.

pnpm-workspace.yaml
```diff
@@ -0,0 +1,3 @@
+packages:
+  - addon
+  - test-app
```
SergeAstapov, line 3: would it cause any issues when workspace is missing?

SergeAstapov (approved):

SergeAstapov:
This is great! Let's do it 🚀 !

.github/workflows/ci-build.yml
NullVoxPopuli: it's been needed until default node is this specific patch, because 18.17 and earlier was successful with webpack, but 18.18, (and 20+) had an issue with libuv, https://github.com/nodejs/node/issues/49911

the specifying specifically, rather than relying on the volta config (or literally any other node version file), but action/setup-node (encapsulated by the wyvox action here) doesn't correctly support package.json as a "node version file" -- I've submitted a PR here: https://github.com/actions/setup-node/issues/864#issuecomment-1752169306 which should automatically propagate once merged and released, but it's been slow going getting anyone to look at (I don't really know who to ping)

addon/config/ember-try.js
NullVoxPopuli: yeah, I'm with ya, it's just, it needs to happen anyway, and as as you state, this PR's been open for ages, I figured since both changes (move to single package monorepo, and pnpm) are pretty small, and internal changes, it's better to combine rather than configure ci.yml entirely differently twice for each of the changes, if split.

addon/package.json
NullVoxPopuli: 100% -- I have to finish https://github.com/release-it-plugins/workspaces/pull/110/ first -- thankfully we have a while before we'll want to release these changes, per: https://github.com/emberjs/ember-test-waiters/issues/458

package.json
NullVoxPopuli: yeep

pnpm-workspace.yaml
NullVoxPopuli: nope

.gitignore
NullVoxPopuli: yeah, I'll address in a follow up

PR #453, opened 2023-05-04
Require ember-auto-import@v2

NullVoxPopuli:

This PR is mostly a no-op be eai2 is required to consume v2 addons.
This will be a breaking change.

I breaking change is a bit of a shame, but this is only a semver communication -- the APIs for modern apps should still be the same, and we can recommend overrides configs for folks to ensure that they only have one copy of `@ember/test-waiters` in their projects.

Moving to a monorepo will allow a good variety of test apps to showcase this.

### DIFF
--- a/package.json
+++ b/package.json
@@ -57,7 +57,7 @@
     "@typescript-eslint/eslint-plugin": "^4.29.2",
     "@typescript-eslint/parser": "^4.30.0",
     "co": "^4.6.0",
-    "ember-cli": "~3.27.0",
+    "ember-cli": "^4.10.0",
     "ember-cli-dependency-checker": "^3.2.0",
     "ember-cli-htmlbars": "^5.7.1",
     "ember-cli-inject-live-reload": "^2.1.0",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -56,8 +57,7 @@
     "@typescript-eslint/eslint-plugin": "^4.29.2",
     "@typescript-eslint/parser": "^4.30.0",
     "co": "^4.6.0",
-    "ember-auto-import": "^1.11.3",
-    "ember-cli": "^4.10.0",
```
NullVoxPopuli, line 60: this shouldn't have been downgraded :thinking:

NullVoxPopuli: fixed

package.json
```diff
@@ -86,7 +86,8 @@
     "qunit": "^2.16.0",
     "release-it": "^14.14.3",
     "release-it-lerna-changelog": "^3.1.0",
-    "typescript": "~4.4.2"
+    "typescript": "~4.4.2",
+    "webpack": "^5.0.0"
```
NullVoxPopuli, line 90: required by ember-auto-import @v2

PR #454, opened 2023-05-04
Remove dual-package emit

NullVoxPopuli:

Historically, this package has emitted two packages: 
- ember-test-waiters
- `@ember/test-waiters`

now only `@ember/test-waiters` remains.

This is a breaking change.
_however_ we should still publish ember-test-waiters so we can ensure that folks can have only one copy of `@ember/test-waiters` in their package graph. `ember-test-waiters` would just be re-exports of `@ember/test-waiters`
Monorepo required for this.


kategengler:
> however we should still publish ember-test-waiters

Do you have this documented as a task somewhere?

NullVoxPopuli:
> Do you have this documented as a task somewhere?


ye, towards the bottom here: https://github.com/emberjs/ember-test-waiters/issues/458

Issue #458, opened 2023-05-05
Quest: Prep for v4: V2 Addon conversion


NullVoxPopuli:
NOTE: the `main` branch will not be releasable during this time, because we need to complete the plan, so that there is minimal/zero disruption to folks' test suites.

Here is a preview of what's in the next major: 
 - https://github.com/emberjs/ember-test-waiters/pull/470

In order of blockers:

- [x] create v3 branch from [this commit](https://github.com/emberjs/ember-test-waiters/commit/790a4f6bbf063196d9de02ff9d3e5df20c36ec51) -- done: https://github.com/emberjs/ember-test-waiters/tree/v3
  - [x] release another v3 version, since the types fix is still unreleased.
      we should keep changes on this branch to the bare minimum, as we don't want to need to sync back to `main` unless we really have to.
      NOTE that the v3 branch is a snapshot of what this repo looked like before everything is this quest issues started happening
- [x] https://github.com/emberjs/ember-test-waiters/pull/454 
- [x] https://github.com/emberjs/ember-test-waiters/pull/453
- [x] https://github.com/emberjs/ember-test-waiters/pull/452

Can happen in any order
- [x] extract test-app(s)
  - [x] create boilerplate for all test apps at once, so the following has more reasonable diffs 
    https://github.com/emberjs/ember-test-waiters/pull/460
  - [x] extract tests from addon to their own test-app 
    https://github.com/emberjs/ember-test-waiters/pull/463
  - [x] separate test-app for ember-concurrency@v3 (it has babel transforms, and the types are different)
    https://github.com/emberjs/ember-test-waiters/pull/465
  - [x] separate test-app for ember-concurrency@v2 (it requires ember-concurrency-ts, which uses a version of `ComputedProperty` which is different from the built-in types) 
    https://github.com/emberjs/ember-test-waiters/pull/464
    - [x] undo this as we don't want to support ecosystem addons in this library
          https://github.com/emberjs/ember-test-waiters/pull/500
  - [x] separate test-app with `ember-fetch@v8`
    - https://github.com/emberjs/ember-test-waiters/pull/471
  - [x] ensure all test-apps are type-checked
    - https://github.com/emberjs/ember-test-waiters/pull/473
- [x] implement solution for [force-highlander](https://github.com/emberjs/ember-test-waiters/pull/455#issuecomment-1743861659)
[...]

PR #461, opened 2023-10-26
Move `WAITER_NAMES` to global context to reduce warning log spew when this addon becomes v2 and duplicates are allowed

NullVoxPopuli:
NOTE: this PR will need to target to the `main` branch as well (which I plan on doing after all the test apps are implemented)

historically, we've only been able to allow one copy of test-waiters in a whole projcet.
this was forced via build-time code in the v1-addon's index.js.
This added maintenance complexity and prevented the conversion to a v2 addon.

With this runtime injection of global state, we can have both the pre-v2 addon
and the v2 addon share the same waiter state.
Also, in `@ember/test-waiters@v3`, we can force the existing build-time code to
take precedecence with the latestVersion of test-waiters v3 (existing behavior),
so there is no risk of old copies of test-waiters not using this runtime highlander code,
because those older copies would not be present in the final build of an app.

-------------

## So what does this PR actually _do_?

- Moves all module state to `globalThis`
  - `getGlobal` already existed, so much of this work was already done
  - the `WAITERS` have already been handled
  - this PR focuses on moving `WAITER_NAMES` to also be on the global context, so we don't create a ton of log spew.
- More specific types to its clearer what the shape of data is for each of these structures
  e.g.:
  ```js
  globalThis: {
    // existing
    [Symbol<'TEST_WAITERS'>]?: Map<WaiterName, Waiter> | undefined;
    // new!
    [Symbol<'TEST_WAITER_NAMES'>]?: Set<string> | undefined;
  }
  ```

-------

## NOTES

- declared compatibility of the 3.x series of this addon include Ember 3.8, which includes IE11 -- so we can't go full native Symbol support yet.

### DIFF
--- a/addon/@ember/test-waiters/build-waiter.ts
+++ b/addon/@ember/test-waiters/build-waiter.ts
@@ -3,34 +3,9 @@ import { Primitive, TestWaiter, TestWaiterDebugInfo, WaiterName } from './';
 import { DEBUG } from '@glimmer/env';
 import { warn } from '@ember/debug';
 import Token from './token';
-import { register, getPrivateData } from './waiter-manager';
+import { register, WAITER_NAMES } from './waiter-manager';
 
 const WAITER_NAME_PATTERN = /^[^:]*:?.*/;
-
-// NOTE: historically, we've only been able to allow one copy of test-waiters in a whole projcet.
-//       this was forced via build-time code in the v1-addon's index.js.
-//       This added maintenance complexity and prevented the conversion to a v2 addon.
-//
-//       With this runtime injection of global state, we can have both the pre-v2 addon
-//       and the v2 addon share the same waiter state.
-//       Also, in `@ember/test-waiters@v3`, we can force the existing build-time code to
-//       take precedecence with the latestVersion of test-waiters v3 (existing behavior),
-//       so there is no risk of old copies of test-waiters not using this runtime highlander code,
-//       because those older copies would not be present in the final build of an app.
-//
-// SAFETY: Types are statically defined,
-//         and we're creating dynamic data on the global scope.
-//         we don't want consumers to know about this data, so we don't
-//         want to extend the type ever. Casting to any is a fine tradeoff.
-getPrivateData().WAITER_NAMES ||= DEBUG ? new Set() : undefined;
-
-function getWaiters(): Set<string> {
-  return getPrivateData().WAITER_NAMES as Set<string>;
-}
-
-export function _resetWaiterNames() {
-  getPrivateData().WAITER_NAMES = new Set();
-}
 
 function getNextToken(): Token {
   return new Token();
@@ -176,10 +151,10 @@ class NoopTestWaiter implements TestWaiter {
  */
 export default function buildWaiter(name: string): TestWaiter {
   if (DEBUG) {
-    warn(`The waiter name '${name}' is already in use`, !getWaiters()!.has(name), {
+    warn(`The waiter name '${name}' is already in use`, !WAITER_NAMES!.has(name), {
       id: '@ember/test-waiters.duplicate-waiter-name',
     });
-    getWaiters()!.add(name);
+    WAITER_NAMES!.add(name);
   }
 
   if (!DEBUG) {
--- a/addon/@ember/test-waiters/build-waiter.ts
+++ b/addon/@ember/test-waiters/build-waiter.ts
@@ -3,13 +3,33 @@ import { Primitive, TestWaiter, TestWaiterDebugInfo, WaiterName } from './';
 import { DEBUG } from '@glimmer/env';
 import { warn } from '@ember/debug';
 import Token from './token';
-import { register } from './waiter-manager';
+import { register, getPrivateData } from './waiter-manager';
 
 const WAITER_NAME_PATTERN = /^[^:]*:?.*/;
-let WAITER_NAMES = DEBUG ? new Set() : undefined;
+
+// NOTE: historically, we've only been able to allow one copy of test-waiters in a whole projcet.
+//       this was forced via build-time code in the v1-addon's index.js.
+//       This added maintenance complexity and prevented the conversion to a v2 addon.
+//
+//       With this runtime injection of global state, we can have both the pre-v2 addon
+//       and the v2 addon share the same waiter state.
+//       Also, in `@ember/test-waiters@v3`, we can force the existing build-time code to
+//       take precedecence with the latestVersion of test-waiters v3 (existing behavior),
+//       so there is no risk of old copies of test-waiters not using this runtime highlander code,
+//       because those older copies would not be present in the final build of an app.
+//
+// SAFETY: Types are statically defined,
+//         and we're creating dynamic data on the global scope.
+//         we don't want consumers to know about this data, so we don't
+//         want to extend the type ever. Casting to any is a fine tradeoff.
+getPrivateData().WAITER_NAMES ||= DEBUG ? new Set() : undefined;
+
+function getWaiters(): Set<string> {
+  return getPrivateData().WAITER_NAMES as Set<string>;
+}
 
 export function _resetWaiterNames() {
-  WAITER_NAMES = new Set();
+  getPrivateData().WAITER_NAMES = new Set();
 }
 
 function getNextToken(): Token {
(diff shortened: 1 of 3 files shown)


addon/@ember/test-waiters/waiter-manager.ts
```diff
@@ -5,20 +5,21 @@ import { registerWaiter } from '@ember/test';
...
+//       in the dep tree. So for now, we can use this private string-based key
+//       that all copies of the `@ember/test-waiters` can use.
+const PRIVATE_GLOBAL_DATA_KEY =
+  '__#private_global_state_for_test_waiters__do_not_access_if_you_want_your_projects_to_keep_working__';
+
 // this ensures that if @ember/test-waiters exists in multiple places in the
 // build output we will still use a single map of waiters (there really should
 // only be one of them, or else `settled` will not work at all)
 const WAITERS: Map<WaiterName, Waiter> = (function () {
-  const HAS_SYMBOL = typeof Symbol !== 'undefined';
```
NullVoxPopuli, line 12: symbols are unique per instantiation, so we can't actually rely on them to be accessible between multiple copies of test-waiters

chancancode: What about https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol/for?

chancancode: Yeah, reading the code more carefully, it seems like it is already doing what you are describing, why isn’t it working?

addon/@ember/test-waiters/build-waiter.ts
```diff
@@ -3,13 +3,33 @@ import { Primitive, TestWaiter, TestWaiterDebugInfo, WaiterName } from './';
 import { DEBUG } from '@glimmer/env';
 import { warn } from '@ember/debug';
 import Token from './token';
-import { register } from './waiter-manager';
+import { register, getPrivateData } from './waiter-manager';
 
 const WAITER_NAME_PATTERN = /^[^:]*:?.*/;
-let WAITER_NAMES = DEBUG ? new Set() : undefined;
```
NullVoxPopuli, line 9: the core of this PR is moving `WAITER_NAMES` from module-state to state that lives on `globalThis` so that it can be shared

addon/@ember/test-waiters/waiter-manager.ts
NullVoxPopuli: thanks, updated!

addon/@ember/test-waiters/waiter-manager.ts
```diff
@@ -1,39 +1,69 @@
...
+// to pin it across their whole dep graph.
+//
+// In the v1 addon version of this library, the build time highlander code will ensure
+// that there is only one copy,
+// And with the v2 addon version of this library, that is where pinning is important,
+// because either:
+// - highlander code will run, and boot out the v2 addon copy
+// - v2 addon has precedence, and the min-version of test-waiters throughout the
+//   dep graph should not preceed the version which this symbol was introduced.
+const PRIVATE_GLOBAL_DATA_KEY = Symbol.for(`@ember/test-waiters' WAITERS`);
```
chancancode, line 23: So I am still confused. The existing code in this file seems to have already made the effort to avoid the problem you described, as indicated by the comment below. Because it effectively uses a global variable, if inadvertently the build code fails to dedup the copies, the multiple copies will still share the same map. This code now changes the symbol name and the structure of the variable, thus defeating the previous attempt and breaking compatibility. Why is that a good idea?

addon/@ember/test-waiters/build-waiter.ts
chancancode: It seems like we don't really need this at all. It appears that the purpose of this is just to keep track of what names has been already been used and issue a warning. But in all cases, immediately the waiter is registered on the `Map` with that string anyway, so it seems like we could have just checked the map for an existing entry in `register`?

addon/@ember/test-waiters/waiter-manager.ts
NullVoxPopuli: no I'm is wrong, the data structures are incompatible. I let you sway me too easily, thinking I've done soething wrong!

the _data stored on the symbol on glabalThis_ is changing:
- instead of a Map its an object with Map and a Set
- this change is because the Set is a logging system to prevent log spew.

So instead, to keep the data structures compatible, I would need to revert this change:
```js
globalThis: {
  [Symbol<PRIVATE_GLOBAL_DATA_KEY>]?: {
    WAITER_NAMES?: Set<string> | undefined;
    WAITERS?: Map<WaiterName, Waiter> | undefined;
  }
}
```
and instead do:
```js
globalThis: {
  [Symbol<TEST_WAITERS>]?: Map<WaiterName, Waiter> | undefined;
  [Symbol<TEST_WAITER_NAMES>]?: Set<string> | undefined;
}
```

Which does add the things on globalThis, but is probably a safer change

addon/@ember/test-waiters/build-waiter.ts
NullVoxPopuli: > Is this theoretical/hypothetical, or were you actually observing the log spew in tests or something? 


This is preparing for switching the library to a v2 addon (See ["The Plan"](https://github.com/emberjs/ember-test-waiters/issues/458)), where we'll have no more highlander code.

Today, with the v1 addon, the WAITER_NAMES stuff is a non-issue because there is only one copy of `@ember/test-waiters` -- that will likely not be the case with native v2 addons.

As for hypothetical -- we're saying that "all waiters, across all parts of the module graph, must have unique names".

chancancode: Ok I was hedging a little bit there and maybe that phrasing confused you. I am saying I didn’t see any evidence that this code is specifically trying to avoid log spew even going back to the original commit. So I am asking where did that concern come from.

Can you read my [comment](https://github.com/emberjs/ember-test-waiters/pull/461#discussion_r1413734731) again? I was specifically saying/clarify “it doesn’t seem like we would have to worry about log spew here based on the usage patterns, is there a specific reason you are concerned?”

I was quite specific about why I don’t think that is an issue here and why I didn’t think this was a specific concern when the code was originally added.

I could definitely be wrong but I am asking where did the concern of log spew came from - was it from deductions from analyzing the code, or from discovering the problem when running the tests, etc.

NullVoxPopuli:
Gonna close this PR because the effort is not worth the benefit.

The changes we want are going to be in the next major anyway, so if folks want the upcoming benefits, they can upgrade and to an overrides/resolution to get the benefit.

The main thing here:
- the list of waiter names is checked eagerly, yet waiters are registered lazily
- the warning is eagerly logged
- we want the map of waiters to be the source of truth

chancancode:
Yeah we discussed on Discord –

* The critical aspect here, which was the bulk of the work in the original PR, is to ensure the actual waiters map is shared. turns out this is already the case, and we can maintain compatibility with the existing version without requiring forcing resolution, etc, which is nice.
* With that out of the way, the secondary thing that this is trying to move off of module state, is the `WAITERS_NAME` set:
  * The purpose of this set is to issue a warning whenever a conflict arise – typically this is either a coding error by an addon author or a module is unexpectedly duplicated. Either way it is a error situation and things will be quite broken. It's not entirely clear to me why it is a warning rather than an error, and it's not very clear to me how much things are expected to degrade "gracefully" in these situations.
  * The reason the conflicting name is a problem is that we store the waiters in a map, using the name as the key, inherently that means there can only be one thing with the same name. Since this is ultimately the reason we care, the natural thing to do would be to warn (or throw, even) when trying to register a second instance with the same name, which would also remove the need for a parallel data structure that tracks the same thing (and we don't currently do a great job keeping the two in-sync – we don't remove the names from the set when unregistering, for example, which probably means the `unregister` API is unusably broken if you intend to re-use the same name).
  * In the current code, moving the warning to the registration side would defer the warning, since `register` isn't called eagerly in `buildWaiter`. It would be slightly sub-optimal and make things slightly more difficult to debug when the problem does happen (the problem = a package consuming `ember-test-waiters` is accidentally duplicated). Personally, I think that is probably acceptable.
  * But it is also not clear that we have a particular good reason to defer adding the waiter to the map in the first place, we can probably just try to move the `register` call out to `buildWaiter` (or in the constructor of the waiter) and see if anything breaks, and solve both problems at once.
  * Overall, I think all of these are fine to change, especially since I consider this warning only advisory in nature. However, there are som slight concerns over doing these kind of changes in a minor release, and since we are going to cut a new major imminently anyway, we may as well do it then.
  * If there is a mix of the new legacy version of ember-test-waiters and the new version in the system, we may not always successfully issue this warning. It is probably acceptable since the warning is advisory, but if we want to, we can also try to address this compatibly in the new major by putting a wrapped `Map` on the global that warn/throw when on the `.set` method when a conflict is detected.

PR #463, opened 2023-12-04
move tests to base-tests app, drop support for 3.8, 3.12, 0.x embroider, and ember-classic

NullVoxPopuli:

This PR moves the base tests out of the addon and in to a real app.
This is because the behavior we're testing isn't so easily isolated to a single library or set of unit tests.

This is phase 1 of test extraction for a series of test-apps that'll be added to ensure that we retain our expected compatibility in the next major.
- dev mode
- production mode
- ember-concurrency@v3 
- ember-concurrency@v2

-----------

Additionally, to make maintenance easier, support (testing in ember-try) has been dropped for 
- ember 3.8
- ember 3.12
- embroider 0.x
- ember-classic
- jquery


.github/workflows/ci-build.yml
```diff
@@ -45,18 +45,14 @@ jobs:
       fail-fast: false
       matrix:
         ember-try-scenario:
-          - ember-lts-3.8
-          - ember-lts-3.12
           - ember-lts-3.16
           - ember-lts-3.20
           - ember-lts-3.24
```
chancancode, line 50: Whether we do it in this PR or not, we probably want to replace the dropped rows with LTS 3.28, 4.4, 4.12. Personally I would be okay with just supporting 3.28+, but I am not sure what additional consideration we have for a "foundational package" like this one.

chancancode (approved):

PR #464, opened 2023-12-04
Move ember-concurrency-specific-tests to the ember-concurrency@v2 test-app

NullVoxPopuli:

This extraction is ocurring so that we can more easily _also_ test ember-concurrency@v3, which is _very hard_ to try to test both v2 and v3 of ember-concurrency in the same app.

Requires: https://github.com/emberjs/ember-test-waiters/pull/463


addon/package.json
```diff
@@ -45,7 +45,7 @@
     "@types/ember-test-helpers": "^1.0.10",
     "@types/ember-testing-helpers": "^0.0.4",
     "@types/ember__test-helpers": "^2.6.1",
-    "@types/qunit": "^2.11.2",
+    "@types/qunit": "^2.19.9",
```
NullVoxPopuli, line 48: this changed because between the different test apps we have different qunit types versions, and it's _just easier_ to have these deps match, than to fix subtrees of the dep graph

test-apps/ember-concurrency-v2/config/ember-try.js
NullVoxPopuli, line 1: same try config as "base-tests"

test-apps/ember-concurrency-v2/tests/unit/wait-for-test.ts
```diff
@@ -0,0 +1,442 @@
+import MockStableError, {
```
NullVoxPopuli, line 1: This test is the same as it was in the base-test app

test-apps/ember-concurrency-v2/tests/unit/wait-for-test.ts
```diff
@@ -0,0 +1,442 @@
...
+// migrating back to mainline off the fork
+// @ts-ignore
+import { task as taskDec } from 'ember-concurrency-decorators';
+import { perform } from 'ember-concurrency-ts';
+// @ts-ignore
+import co from 'co';
+
+// @ember/test-waiters is still a v1 addon and is too weird
+// to have in-repo types working correctly.
+// @ts-ignore
```
NullVoxPopuli, line 26: ts-ignores can be cleaned up once test-waiters is a v2 addon and actually has types we can consume

chancancode (approved):
caveat: I didn't look at the moved tests to make sure they are moved correctly but overall everything seems to make sense

PR #466, opened 2023-12-04
Add support for 3.28, 4.12, 5.4

NullVoxPopuli:

_Preparing for the v4 release_.

### DIFF
--- a/.github/workflows/ci-build.yml
+++ b/.github/workflows/ci-build.yml
@@ -51,6 +51,9 @@ jobs:
           - ember-lts-3.16
           - ember-lts-3.20
           - ember-lts-3.24
+          - ember-lts-3.28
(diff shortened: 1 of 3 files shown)


chancancode (approved):

chancancode:
Seems good, but do we want to explicitly test 4.4, etc, (or drop 3.16, etc from the matrix), to be consistent?

NullVoxPopuli:
> or drop 3.16, etc from the matrix

I'd learn towards this, but I'd want it to be explicitly called out in the changelog (changelog entrie = PR title).

I also want to make some more progress on testing, etc to see what maximum support is.

Because if it's easy to support 3.16, we could omit 3.20 and 3.24 from the matrix instead

PR #473, opened 2024-05-04
Drop support for TS < v5 -- Setup CI for type-checking all test apps

NullVoxPopuli:
Changes
- use ember-source built in types (DT types are best effort, and not great for modern use as they are not compatible with ember-data (ember-data is compatible with the official types now))
- update the base devDep version of ember-source to support ^
- Add typescript test matrix
- Drop support for typescript < 5
  - This is because `@tsconfig/ember` is only compatible with TS v5, due to having `verbatimModuleSyntax` specified 
- Test against all TS 5.x versions
  - NOTE: ember-concurrency@v2 is not compatible with the built-in ember-source types. It thinks ComputedProperty is generic (which was true from the DT types). So I've removed type-checking from the ember-concurrency test-app for now. In a future PR I think it makes sense to delete the ember-concurrency test-app entirely, because looking at the implementation of `@waitFor` we only _happen_ to work with some ember-concurrency versions do to being compatible with thennables and generators -- we could probably test that stuff in isolation without ember-concurrency. 
- Add overrides for certain `@glimmer/*` packages due to: https://github.com/glimmerjs/glimmer.js/issues/413
  - also because older versions' types are not compatible with the latest stricter tsconfig from `@tsconfig/ember`


test-apps/ember-concurrency-v2/tests/unit/wait-for-test.ts
NullVoxPopuli, line 1: To address in a separate PR: getting linting (ci) agreeing with local config, upgrading lint configs, etc (prettier 3), etc

gitKrystan (approved):

PR #474, opened 2024-05-04
Drop support for Typescript < 4.9

NullVoxPopuli:
We want to use ember-source's built in types for internal type checking, and we can't do that with TS < v4.9

TS v4.9 was released November of 2022

### DIFF
--- a/addon/package.json
+++ b/addon/package.json
@@ -76,7 +76,7 @@
     "npm-run-all": "^4.1.5",
     "prettier": "^2.5.1",
     "qunit": "^2.20.0",
-    "typescript": "~4.4.2",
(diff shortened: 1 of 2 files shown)


kategengler:
Ember itself still supports 4.9 https://github.com/emberjs/ember.js/blob/main/.github/workflows/ci.yml#L57

NullVoxPopuli:
legit, still way better than 4.4 I'll update

kategengler (approved):

kategengler:
If you follow TS Semver, like Ember, a change in TS versions doesn't need to be breaking iirc https://www.semver-ts.org/

PR #476, opened 2024-05-05
Convert to V2 Addon, drop support for ember-source < v4

NullVoxPopuli:
Applies https://github.com/embroider-build/addon-blueprint
+
some updates from ember-cli 6.1 (better eslint)

Marked as breaking because 
- in v1 apps and addons: ember-auto-import is required to import `@ember/test-waiters`
-  minimum ember-source is 4.0 -- due to: `Error: @ember/test-helpers needs to depend on ember-auto-import in order to use @ember/test-waiters` 
   - `@ember/test-helpers` v3 has ember-auto-import
     - only supports ember-source v4 and above 
   - `@ember/test-helpers` v2 does _not_ have ember-auto-import
     - supports ember-source earlier than v4, but this version _cannot_ be used with the v2 addon version of `@ember/test-waiters`  

Comparing with v3:
- https://www.npmjs.com/package/@ember/test-waiters?activeTab=code
- full diff: https://github.com/emberjs/ember-test-waiters/compare/v3.1.0...v2-addon
- upcoming release notes: https://github.com/emberjs/ember-test-waiters/pull/470


addon/types/custom-ember-debug.d.ts
NullVoxPopuli, line 1: no longer needed, due to using build in types

addon/types/dummy/index.d.ts
NullVoxPopuli, line 1: no longer needed, due to using build in types

addon/types/qunit/index.d.ts
NullVoxPopuli, line 1: no longer needed, due to using build in types

test-apps/base-tests/config/ember-try.js
```diff
@@ -2,7 +2,7 @@
 
 const getChannelURL = require('ember-source-channel-url');
 
-module.exports = async function() {
+module.exports = async function () {
```
NullVoxPopuli, line 5: new prettier settings (not changeable)

test-apps/base-tests/package.json
```diff
@@ -63,17 +64,17 @@
...
+    "ember-modifier": "^4.2.0",
+    "ember-page-title": "^8.2.3",
+    "ember-qunit": "^8.1.1",
+    "ember-resolver": "^13.1.0",
     "ember-source": "~5.3.0",
     "ember-source-channel-url": "^3.0.0",
     "ember-template-lint": "^6.0.0",
     "ember-try": "^3.0.0",
     "ember-welcome-page": "^7.0.2",
-    "eslint": "^9.15.0",
```
NullVoxPopuli, line 76: this was a mistake on main/master since it's been too hard to keep CI green throughout this process (of de-tangling some goofy v1 magic)

.github/workflows/ci-build.yml
```diff
@@ -4,7 +4,6 @@ on:
   push:
     branches:
       - master
-      - 'v*'
```
NullVoxPopuli, line 7: we're not releasing on tag-pushes as we have a mass migration to release-plan which tags as a part of releasing, rather than releasing because of a tag

.github/workflows/ci-build.yml
```diff
@@ -19,47 +18,30 @@ jobs:
...
       - uses: wyvox/action-setup-pnpm@v3
       # type checking requires build
-      - run: pnpm prepack
-        working-directory: addon
+      - run: pnpm build
       - run: pnpm lint
 
   test:
     timeout-minutes: 10
-    name: "Tests: ${{ matrix.EMBROIDER && 'embroider' || 'classic' }}"
```
NullVoxPopuli, line 28: there is no longer any build behavior, so we aren't testing embroider and classic in a matrix -- we expect that embroider handles any behaviors used by this library

.github/workflows/ci-build.yml
```diff
@@ -108,16 +90,12 @@ jobs:
           - test-apps/base-tests
         ember-try-scenario:
```
NullVoxPopuli, line 91: to cut down on CI noise, this is now just testing the ranges of support,

3.16 - minimum
x.latest - last LTS for a major, X
the 3 variable embers

addon/.eslintrc.cjs
NullVoxPopuli, line 1: eslint 9 + flat config isn't available until ember-cli 6.1

addon/.npmignore
NullVoxPopuli, line 1: replaced by package.json#files

addon/package.json
```diff
@@ -86,9 +63,37 @@
...
+    "dist",
+    "src",
+    "addon-main.cjs"
+  ],
+  "exports": {
+    ".": {
+      "types": "./declarations/index.d.ts",
+      "default": "./dist/index.js"
+    },
+    "./__private__/types": {
```
NullVoxPopuli, line 77: used by tests only. there were no import restrictions in the prior version of test-waiters.

addon/src/waiter-manager.ts
```diff
@@ -1,4 +1,4 @@
-import { PendingWaiterState, Waiter, WaiterName } from './types';
```
NullVoxPopuli, line 1: using type imports is important for not shipping too much stuff to the runtime (especially stuff that doesn't exist)

package.json
```diff
@@ -7,31 +7,24 @@
...
     "lint": "pnpm --filter '*' lint",
     "lint:fix": "pnpm --filter '*' lint:fix",
-    "start": "concurrently 'npm:start:*' --restart-after 5000 --prefix-colors cyan,white,yellow",
+    "start": "concurrently \"pnpm:start:*\" --restart-after 5000 --prefix-colors cyan,white,yellow",
     "start:addon": "pnpm --filter @ember/test-waiters start -- --no-watch.clearScreen",
     "start:test-app": "pnpm --filter test-app start",
     "test": "pnpm --filter '*' test"
   },
   "devDependencies": {
+    "@glint/core": "^1.5.0",
```
NullVoxPopuli, line 19: because of vscode reasons, some LSP/Tooling dependencies need to be added to the root package.json.

No CLI tool cares that these are here

.github/workflows/ci-build.yml
```diff
@@ -19,47 +18,41 @@ jobs:
       - uses: actions/checkout@v4
       - uses: wyvox/action-setup-pnpm@v3
```
SergeAstapov, line 19: why not sync up with blueprint to use `pnpm/action-setup` and `actions/setup-node`?

.github/workflows/ci-build.yml
SergeAstapov: @NullVoxPopuli looks like some kind of mistake as now it only has 4.0, 4.12 and 5.12, no more 3.16 and no 3.28?

addon/package.json
```diff
@@ -86,9 +63,37 @@
     "access": "public",
     "registry": "https://registry.npmjs.org"
   },
+  "files": [
+    "declarations",
+    "dist",
+    "src",
```
SergeAstapov, line 69: @NullVoxPopuli why would we need src to be published?

addon/package.json
```diff
@@ -11,73 +11,50 @@
...
+    "eslint": "^9.16.0",
     "eslint-config-prettier": "^9.1.0",
     "eslint-plugin-ember": "^12.3.3",
+    "eslint-plugin-n": "^17.14.0",
     "eslint-plugin-node": "^11.1.0",
     "eslint-plugin-prettier": "^5.2.1",
-    "globals": "^15.12.0",
-    "loader.js": "^4.7.0",
+    "execa": "^9.5.1",
+    "fix-bad-declaration-output": "^1.1.4",
```
SergeAstapov, line 49: is this actually needed?

addon/src/build-waiter.ts
```diff
@@ -128,7 +128,7 @@ class NoopTestWaiter implements TestWaiter {
 /**
  * Builds and returns a test waiter. The type of the
  * returned waiter is dependent on whether the app or
- * addon is in `isDevopingApp()` mode or not.
```
SergeAstapov, line 131: I may like to Dev Oping, more like Dev Oopsing lol

test-apps/base-tests/tests/unit/build-waiter-test.ts
```diff
@@ -13,7 +13,7 @@ import { module, test } from 'qunit';
 
 // @ts-ignore
 import { Promise } from 'rsvp';
-import Token from '@ember/test-waiters/token';
+import Token from '@ember/test-waiters/__private__/token';
```
SergeAstapov, line 16: how does this extra `/__private__/` works?

SergeAstapov (approved):

.github/workflows/ci-build.yml
NullVoxPopuli: less to think about, tbh.
also also back when I made action-setup-pnpm, both setup-node and pnpm/action-setup did not infer either of their versions from the correct places. 
_Today_ the difference is (per job) 1 line vs ~ 4, I think, not too bad

.github/workflows/ci-build.yml
NullVoxPopuli: this comment is out of date actually -- we cannot support ember < v4

addon/package.json
NullVoxPopuli: so go-to-definition can go to source instead of declarations (requires declaration maps, iirc)

addon/package.json
NullVoxPopuli: it's a default thing I use in v2 addons, because the `/// <reference` thing with peers is too annoying to fix otherwise if it comes up

addon/src/build-waiter.ts
NullVoxPopuli: lol

test-apps/base-tests/tests/unit/build-waiter-test.ts
NullVoxPopuli: https://github.com/emberjs/ember-test-waiters/pull/476/files#diff-6d28e496920bb420e41d13a83301a26ed6f19c4cec4dac3723042b1253d87181R77

addon/package.json
SergeAstapov: ok, so it's just in case here rather actually needed?

NullVoxPopuli: Idk. It is used in the rollup config.

It's just not a problem i want to run in to, since TS is very against properly working peers - there isn't a better path forward:(

SergeAstapov: thanks for explanation! IMO it's ok to keep it as long as any future reader can easily understand/use it which should be the case here

SergeAstapov (approved):

ef4:
> in all addons, declaring @ember/test-waiters in package.json is required to import from it

I think that this is not supposed to be true and if it's true it's a bug in ember-auto-import. We try to maintain compatibility so that all v2 addons are always available to every package, in keeping with the classic behavior.

NullVoxPopuli:
so far, in the auto-import test suite: I'm unable to create a rero 🤔 https://github.com/embroider-build/ember-auto-import/pull/650/files

maybe everything is fine then!

NullVoxPopuli:
removed the second bullet -- the only requirement for v1 addons is that they have ember-auto-import -- as long as they do, folks should be able to use the new/upcoming version of `@ember/test-waiters`

addon/package.json
NullVoxPopuli: yea, in the rollup config I link to some discussion on the TS repo about peer-provided types

SergeAstapov (approved):

PR #513, opened 2025-02-17
Implement functionally for "deprecate fetch" RFC

NullVoxPopuli:
- https://github.com/emberjs/rfcs/pull/1065

### DIFF
--- a/addon/src/wait-for-fetch.ts
+++ b/addon/src/wait-for-fetch.ts
@@ -12,7 +12,6 @@ export async function waitForFetch(fetchPromise: ReturnType<typeof fetch>) {
         ['json', 'text', 'arrayBuffer', 'blob', 'formData', 'bytes'].includes(prop)
       ) {
         return (...args: unknown[]) => {
-          console.log('invoke', ...args);
           return waitForPromise(original.call(target, ...args));
         };
       }
--- a/addon/src/wait-for-fetch.ts
+++ b/addon/src/wait-for-fetch.ts
@@ -1,10 +1,7 @@
 import { default as waitForPromise } from './wait-for-promise';
 
 export async function waitForFetch(fetchPromise: ReturnType<typeof fetch>) {
-  // eslint-disable-next-line @typescript-eslint/no-floating-promises
-  waitForPromise(fetchPromise);
-
-  const response = await fetchPromise;
+  const response = await waitForPromise(fetchPromise);
 
   return new Proxy(response, {
     get(target, prop, receiver) {
--- a/addon/src/wait-for-fetch.ts
+++ b/addon/src/wait-for-fetch.ts
@@ -7,13 +7,12 @@ export async function waitForFetch(fetchPromise: ReturnType<typeof fetch>) {
     get(target, prop, receiver) {
       const original = Reflect.get(target, prop, receiver);
 
-      console.log(...arguments);
       if (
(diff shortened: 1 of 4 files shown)


addon/src/wait-for-fetch.ts
```diff
@@ -0,0 +1,27 @@
+import { default as waitForPromise } from './wait-for-promise';
+
+export async function waitForFetch(fetchPromise: ReturnType<typeof fetch>) {
+  // eslint-disable-next-line @typescript-eslint/no-floating-promises
+  waitForPromise(fetchPromise);
+
+  const response = await fetchPromise;
```
ef4, line 7:
```suggestion
  const response = await waitForPromise(fetchPromise);
```

addon/src/wait-for-fetch.ts
```diff
@@ -0,0 +1,27 @@
...
+      const original = Reflect.get(target, prop, receiver);
+
+      if (
+        typeof prop === 'string' &&
+        ['json', 'text', 'arrayBuffer', 'blob', 'formData'].includes(prop)
+      ) {
+        return (...args: unknown[]) => {
+          const parsePromise = original.call(target, ...args);
+
+          return waitForPromise(parsePromise);
```
ef4, line 20:
```suggestion
          return waitForPromise(original.call(target, ...args));
```

addon/src/wait-for-fetch.ts
```diff
@@ -0,0 +1,27 @@
...
+
+  const response = await fetchPromise;
+
+  return new Proxy(response, {
+    get(target, prop, receiver) {
+      const original = Reflect.get(target, prop, receiver);
+
+      if (
+        typeof prop === 'string' &&
+        ['json', 'text', 'arrayBuffer', 'blob', 'formData'].includes(prop)
```
ef4, line 15:
```suggestion
        ['json', 'text', 'arrayBuffer', 'blob', 'formData', 'bytes'].includes(prop)
```

addon/src/wait-for-fetch.ts
ef4: The eslint-disable weirded me out, when there's no need to write it this way anyway.

addon/src/wait-for-fetch.ts
ef4: I found one more async method in the MDN docs.

mansona:
@NullVoxPopuli can you change the title of this PR to `Implement functionally for "deprecate fetch" RFC` so it will make better sense in the changelog. I would have done it myself but it seems like I don't have enough permissions 🤔

NullVoxPopuli:
ope, yes! thanks

addon/src/wait-for-fetch.ts
```diff
@@ -0,0 +1,23 @@
...
+  return new Proxy(response, {
+    get(target, prop, receiver) {
+      const original = Reflect.get(target, prop, receiver);
+
+      if (
+        typeof prop === 'string' &&
+        ['json', 'text', 'arrayBuffer', 'blob', 'formData', 'bytes'].includes(prop)
+      ) {
+        return (...args: unknown[]) => {
+          console.log('invoke', ...args);
```
ef4, line 15: console.log

NullVoxPopuli:
@ef4 resolved

PR #519, opened 2025-06-13
fix: proxy more parts of fetch Response

### LINKED ISSUE #518: improve waitForFetch proxy
Hello!

TLDR; I think we need to update the waitForFetch() function to proxy more of the Response. After updating from pretender -> MSW I noticed a few issues, likely due to MSW using a different Response to what pretender provides

---

After switching to MSW, I noticed that waitForFetch's Response Proxy started throwing errors on some of the calls made against the Response. Specifically, the `Reflect.get()` call would throw "Illegal invocation" error, when `ok`, `status`, `statusText` etc were asked for - any _properties_ of Response. Functions calls like `json` etc remained fine.

https://github.com/emberjs/ember-test-waiters/blob/db982ddd4feafad5c7f506f80ef87d7633846e4b/addon/src/wait-for-fetch.ts#L6-L16

Some research into the error suggests that this is something to do with the `receiver` arg there to Reflect.get(). To fix this, we just make sure to avoid it for property calls like `ok` etc.

Additionally, for function calls like `clone()` that aren't async, we need to call them (but no need to wrap in waitForPromise for those).

My patch looks like this currently, and it's working great. I'll probably turn this into a PR

```diff
   const response = await waitForPromise(fetchPromise);
   return new Proxy(response, {
     get(target, prop, receiver) {
+      if (
+        typeof prop === 'string' &&
+        ['ok', 'status', 'statusText', 'bodyUsed', 'headers', 'redirected', 'type', 'url'].includes(prop)
+      ) {
+        return target[prop];
+      }
       const original = Reflect.get(target, prop, receiver);
       if (typeof prop === 'string' && ['json', 'text', 'arrayBuffer', 'blob', 'formData', 'bytes'].includes(prop)) {
         return (...args) => {
           return waitForPromise(original.call(target, ...args));
         };
       }
+      if (typeof prop === 'string' && ['clone'].includes(prop)) {
+        return (...args) => {
+          return original.call(target, ...args);
+        }
+      }
       return original;
```

Techn1x:
resolves https://github.com/emberjs/ember-test-waiters/issues/518

### DIFF
--- a/addon/src/wait-for-fetch.ts
+++ b/addon/src/wait-for-fetch.ts
@@ -1,18 +1,33 @@
 import { default as waitForPromise } from './wait-for-promise.ts';
 
-const properties = [
-  'ok',
-  'status',
-  'statusText',
+const props = [
+  'body',
   'bodyUsed',
   'headers',
+  'ok',
   'redirected',
+  'status',
+  'statusText',
   'type',
   'url',
 ] as const satisfies (keyof Response)[];
-type ResponseProp = (typeof properties)[number];
+type ResponseProp = (typeof props)[number];
 function isResponseProperty(maybeProp: string): maybeProp is ResponseProp {
-  return properties.some((prop) => maybeProp === prop);
+  return props.some((prop) => maybeProp === prop);
+}
+
+const fns = [
+  'arrayBuffer',
+  'blob',
+  'bytes',
+  'clone',
+  'formData',
+  'json',
+  'text',
+] as const satisfies (keyof Response)[];
+type ResponseFn = (typeof fns)[number];
+function isResponseFn(maybeFn: string): maybeFn is ResponseFn {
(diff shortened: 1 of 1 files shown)


NullVoxPopuli:
Sorry it took me so long to get to this -- I forgot :( 

Can you address the type issue? thanks!!

Techn1x:
I've gone a bit further than asked and;
* fixed the type issue, and ensured that the property names we're using are indeed properties of Response class, by using type guards
* added some code comments. I tend to over-comment so please let me know if I should reduce them.
* added response `body` property to list of properties

Probably a little more challenging to review now, sorry!

@NullVoxPopuli

Techn1x:
FYI @mfeckie - we can remove some ember/test-waiters pnpm patches we have with this update in the 4.1.1 release. Shoot me a message if it causes trouble

PR #525, opened 2026-08-10
[SPIKE] waitersSettled(): await waiter completion instead of polling for it

NullVoxPopuli-ai-agent:
Companion to emberjs/ember-test-helpers#1574, exploring the RFC 957 test story (context: emberjs/ember.js#21520). Draft for discussion.

## The idea

Waiters already know the moment each operation completes, but consumers can only find out by polling `hasPendingWaiters()` on a timer. So:

- `beginAsync` creates a promise for that operation,
- `endAsync` resolves it and drops it from the pending map — which therefore only ever holds work still in flight,
- `waitersSettled()` composes the promises of whatever is currently tracked,
- and `Waiter` gains an optional `settled()` so each waiter answers for its own operations.

`NoopTestWaiter` (production) resolves immediately; nothing changes for non-test builds.

## Two honest limits, both reflected in the API

**Operations begun after the call are not included.** Settling can start more work, so `waitersSettled()` is a snapshot, and a caller that needs a true fixpoint re-checks. (The companion `settled()` loops.)

**A `Waiter` implemented directly against the interface may not implement `settled()`** — its `waitUntil` is pull-only and nothing obliges it to announce anything. Rather than claiming a quiet it cannot verify, `waitersSettled()` returns a **never-resolving** promise while such a waiter is pending, so callers race it against their own fallback tick and the pull-only waiter still works.

## Does it actually do the work?

Measured, not assumed. Instrumenting the companion `settled()` to record which side of the race won, in an app whose suite exercises workers, rendering, and compilation: **91 of 92 iterations decided by the promises**, 1 by the fallback.

That measurement also surfaced a sharp edge worth recording: the fallback tick must be slower than a frame. At 10ms it beat *frame-paced render ticks* to the race and decided 30 of 117 iterations, costing an extra loop pass each time; at 50ms, 1 of 92.

## Validation

- This repo's `base-tests` app: **56/56**.
- `@ember/test-helpers` test-app against a linked build of this branch: **553/553**; against published `@ember/test-waiters` (no `waitersSettled`, degraded to the fallback tick): **553/553**.
- Against the runloop-less ember spike: limber's full chrome matrix green, 7/7 repeat runs on the suite that exposed the macrotask-confirm flake.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/addon/src/build-waiter.ts
+++ b/addon/src/build-waiter.ts
@@ -15,6 +15,21 @@ function getNextToken(): Token {
   return new Token();
 }
 
+/**
+ * A promise that carries its own resolver, so a collection of pending
+ * operations holds nothing but the promises themselves.
+ */
+type ResolvablePromise = Promise<void> & { resolve: () => void };
+
+function resolvablePromise(): ResolvablePromise {
+  let resolve!: () => void;
+  const promise = new Promise<void>((r) => (resolve = r)) as ResolvablePromise;
+
+  promise.resolve = resolve;
+
+  return promise;
+}
+
 class TestWaiterImpl<T extends object | Primitive = Token> implements TestWaiter<T> {
   public name: WaiterName;
   private nextToken: () => T;
@@ -23,6 +38,14 @@ class TestWaiterImpl<T extends object | Primitive = Token> implements TestWaiter
   items = new Map<T, TestWaiterDebugInfo>();
   completedOperationsForTokens = new WeakMap<Token, boolean>();
   completedOperationsForPrimitives = new Map<Primitive, boolean>();
+
+  /**
+   * The completion promise for each pending operation, keyed by token so
+   * `endAsync` can settle its own. Values are the promises themselves,
+   * each carrying its resolver, and an entry is dropped as it resolves --
+   * so this only ever holds operations still in flight.
+   */
+  private pendingPromises = new Map<T, ResolvablePromise>();
 
   constructor(name: WaiterName, nextToken?: () => T) {
     this.name = name;
@@ -47,6 +70,8 @@ class TestWaiterImpl<T extends object | Primitive = Token> implements TestWaiter
       label,
     });
 
+    this.pendingPromises.set(token, resolvablePromise());
+
     return token;
   }
 
@@ -64,6 +89,13 @@ class TestWaiterImpl<T extends object | Primitive = Token> implements TestWaiter
     // Mark when a waiter operation has completed so we can distinguish
     // whether endAsync is being called before a prior beginAsync call above.
     this._getCompletedOperations(token).set(token, true);
+
+    const pending = this.pendingPromises.get(token);
+
+    if (pending !== undefined) {
+      this.pendingPromises.delete(token);
+      pending.resolve();
+    }
   }
 
   waitUntil(): boolean {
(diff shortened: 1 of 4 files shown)


addon/src/build-waiter.ts
```diff
@@ -24,6 +24,13 @@ class TestWaiterImpl<T extends object | Primitive = Token> implements TestWaiter
   completedOperationsForTokens = new WeakMap<Token, boolean>();
   completedOperationsForPrimitives = new Map<Primitive, boolean>();
 
+  /**
+   * The completion promise for each pending operation, resolved by the
+   * `endAsync` that ends it and dropped at the same time -- so this only
+   * ever holds operations still in flight.
+   */
+  private pendingOperations = new Map<T, { promise: Promise<void>; resolve: () => void }>();
```
NullVoxPopuli: this should be a Set containing only promises.

Then, 

settled can still Promise.all(the set's values)

addon/src/build-waiter.ts
```diff
@@ -80,8 +99,20 @@ class TestWaiterImpl<T extends object | Primitive = Token> implements TestWaiter
     return result;
   }
 
+  settled(): Promise<void> {
+    const operations = [...this.pendingOperations.values()].map((operation) => operation.promise);
```
NullVoxPopuli: never ever ever array splat

addon/src/build-waiter.ts
```diff
@@ -122,6 +153,10 @@ class NoopTestWaiter implements TestWaiter {
     return [];
   }
 
+  settled(): Promise<void> {
+    return Promise.resolve();
```
NullVoxPopuli: why? this is pointless, no?

addon/src/build-waiter.ts
```diff
@@ -80,8 +112,20 @@ class TestWaiterImpl<T extends object | Primitive = Token> implements TestWaiter
     return result;
   }
 
+  settled(): Promise<void> {
+    return Promise.all(this.pendingPromises.values()).then(() => {});
```
NullVoxPopuli: why did you add this `then`? its unneeded

addon/src/build-waiter.ts
NullVoxPopuli-ai-agent: Done — the collection now holds only promises. Each one carries its own resolver (`promise.resolve = resolve` at creation), so there are no `{promise, resolve}` wrapper objects and `settled()` reads promises directly.

One deviation I want to flag rather than hide: it's a `Map<Token, ResolvablePromise>` and not a `Set`, because `endAsync(token)` has to settle *that* operation, which needs a lookup key. `values()` is the set of promises, so `settled()` is `Promise.all(this.pendingPromises.values())`. A literal `Set<Promise>` would need a second token→promise structure alongside it. Happy to switch if you'd rather have the Set and eat the extra map.

addon/src/build-waiter.ts
NullVoxPopuli-ai-agent: Fixed. `Promise.all` takes any iterable, so it's now `Promise.all(this.pendingPromises.values())` — the splat and the `.map` are both gone.

addon/src/build-waiter.ts
NullVoxPopuli-ai-agent: You're right — removed. `NoopTestWaiter` never calls `register()`, so it's never in `getWaiters()` and `waitersSettled()` would never reach it. `settled()` is optional on the interface, so nothing needs the stub.

addon/src/build-waiter.ts
NullVoxPopuli-ai-agent: Removed. It was only there to make the signature `Promise<void>` instead of `Promise<void[]>` — a pointless allocation and extra microtask for a value nobody reads. `settled()` and `waitersSettled()` now return `Promise<unknown>` and hand back `Promise.all(...)` directly.

